PR #9, opened 2018-07-20
Enable GLRenderer.

suyambulingamrm:
This CL
i. Adds method to retrieve texture backed SkImage from bitmap pixels.
ii. Enable GL Renderer to composite bitmap raster data.

### DIFF
--- a/cc/resources/display_resource_provider.cc
+++ b/cc/resources/display_resource_provider.cc
@@ -10,6 +10,10 @@
 #include "gpu/command_buffer/client/gles2_interface.h"
 #include "third_party/skia/include/gpu/GrBackendSurface.h"
 
+#if defined(CASTANETS)
+#include "skia/ext/texture_handle.h"
+#endif
+
 using gpu::gles2::GLES2Interface;
 
 namespace cc {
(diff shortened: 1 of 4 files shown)


cc/resources/display_resource_provider.cc
```diff
@@ -28,6 +29,7 @@ DisplayResourceProvider::DisplayResourceProvider(
 DisplayResourceProvider::~DisplayResourceProvider() {
   while (!children_.empty())
     DestroyChildInternal(children_.begin(), FOR_SHUTDOWN);
+  resource_sk_image_.clear();
```
Sunwoo-Nam: How do you think whether a Castanets flag is needed for every modification or not?

components/viz/service/display/gl_renderer.cc
```diff
@@ -600,7 +600,9 @@ void GLRenderer::DoDrawQuad(const DrawQuad* quad,
       DrawSolidColorQuad(SolidColorDrawQuad::MaterialCast(quad), clip_region);
       break;
     case DrawQuad::STREAM_VIDEO_CONTENT:
+#if !defined(CASTANETS)
       DrawStreamVideoQuad(StreamVideoDrawQuad::MaterialCast(quad), clip_region);
```
Sunwoo-Nam: Why did you ignore video quad rendering?

cc/resources/display_resource_provider.cc
suyambulingamrm: I thought, the changes in ScopedReadLockSkImage is generic. So didnt add build flag.
I have added build flags accordingly in latest patch set.

components/viz/service/display/gl_renderer.cc
suyambulingamrm: Currently, StreamVideoQuad and YUV quads are not encountered in the supported scenarios.
Also it needs some verification to make the changes for these quads.
So I have added a TODO for now. Once these quads are supported/needed we will add a fixup.
Should that be fine ?

PR #10, opened 2018-07-23
Fix utility process error issue.

Sunwoo-Nam:
These commits avoid running utility features to fix utility process error


suyambulingamrm:
LGTM.
The changes are needed as launch issue is observed on some environments.

PR #12, opened 2018-07-25
Discovery client to run renderer process on remote

djmixkim:
This patch request discovery client to run renderer
process on the remote node via D-Bus.

Original author: yh106.jung@samsung.com

Signed-off-by: DongJun Kim <djmix.kim@samsung.com>


djmixkim:
Dear @Sunwoo-Nam 
I talked about this with @jungyoungha ,
IMO, Castanets is independent and executable project by itself.
We had better that do not have dependency with other project like as service discovery.

So we will update new patch-set as below
- Runtime-flag for service discovery (Default : disable)
- Update README file for Service Discovery Manager

Could you merge this patch for now?

PR #26, opened 2018-10-17
Enable Media Element to run using shared bitmap transmission.

Sunwoo-Nam:
Share a video texture using TCP socket. If a video texture is created
on renderer process, it is sent in shared bitmap format.
And adjust frame time for the rendering frame decision algorithm.

Signed-off-by: Sunwoo Nam <jegalzz88@gmail.com>


Sunwoo-Nam:
I am going to make the fixup patch for the remaining issues

PR #37, opened 2018-11-27
Add proxy configuration support to the network service

joonghunpark:
These CLs' purpose is to add proxy configuration to the network service.
Without these changes, chromium can't load the external resources in proxy circumstances.


venumusham:
lgtm

PR #38, opened 2018-11-28
Disable scrambled msg id feature for castanets.

suyambulingamrm:
Scrambled msg id feature is supported only for linux platform in chromium.
In order to make communication between different platforms like tizen,
android this has to be disabled.


venumusham:
lgtm

PR #39, opened 2018-12-05
Use Hash for IPC Message instead of __LINE__

is46-kim:
Unique ID of IPC Message is given by __LINE__. It caused a IPC message compatibility issue between different chromium versions. The IPC Unique ID consists of class ID and line number. Class IDs are defined in ipc_message_start.h file, and the number is 127 or less.
This patch resuces length of the class ID and uses the hash of IPC Message instead of line number(__LINE__) to give unique IDs.

Before : [16bit: CLASS ID][16bit: __LINE__]
After  : [ 7bit: CLASS ID][25bit: HASH of Message]


venumusham:
lgtm

PR #50, opened 2018-12-28
Fix build break for Android build.

Sunwoo-Nam:
fixup! Network Service on renderer.

Fix build break for Android build.


suyambulingamrm:
@Sunwoo, windows build also breaks because of this.. 
Shall we add the same on _win.cc file as well.. ?

PR #51, opened 2019-01-02
Adjust web preferences to connect to the Android target.

Sunwoo-Nam:
When create a view, deserialize web preferences on renderer side.
But some variables are wrapped in OS_ANDROID macro and it causes failure.
Hence, web preferences need to be synchronized for inter-platform connection.


yg48:
+1 :)

suyambulingamrm:
+1.

PR #53, opened 2019-01-31
Enable Network Service to run on renderer device without dependency on Network File System.

venumusham:
Cleanup debugging code which prints resource contents
on output.

Signed-off-by: venu.musham <venu.musham@samsung.com>

### DIFF
--- a/mojo/edk/system/node_controller.cc
+++ b/mojo/edk/system/node_controller.cc
@@ -1120,18 +1120,26 @@ void NodeController::OnIntroduce(const ports::NodeName& from_node,
     return;
   }
 #if defined(CASTANETS)
-  std::string process_type_str =
-      base::CommandLine::ForCurrentProcess()->GetSwitchValueASCII("type");
-  ScopedPlatformHandle handle;
-  if (process_type_str == "utility")
-    handle = mojo::edk::CreateTCPServerHandle(mojo::edk::kCastanetsNonBrokerPort);
-  else
-    handle = mojo::edk::CreateTCPClientHandle(mojo::edk::kCastanetsNonBrokerPort);
+  scoped_refptr<NodeChannel> channel;
+  if (base::Castanets::IsEnabled()) {
+    std::string process_type_str =
+        base::CommandLine::ForCurrentProcess()->GetSwitchValueASCII("type");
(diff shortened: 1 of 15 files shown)


mojo/edk/system/node_controller.cc
```diff
@@ -370,7 +376,9 @@ void NodeController::SendBrokerClientInvitationOnIOThread(
 
 #if !defined(OS_MACOSX) && !defined(OS_NACL) && !defined(OS_FUCHSIA)
 #if defined(CASTANETS)
-  ScopedPlatformHandle server_handle = mojo::edk::CreateTCPServerHandle(mojo::edk::kCastanetsBrokerPort);
+  static int port = mojo::edk::kCastanetsBrokerPortUtility;
```
suyambulingamrm: I think this will cause issue when opening second tab.

joonghunpark:
As a first step, this change should be rebased.
And then, we need to think about file:// protocol file loading scenario.
Currently, when file protocol url is entered in url bar, the file located in device running network service process is loaded.

I wonder whether it is possible to change the file protocol file's path to include browser process running device's ip before start loading, so url loader could load through it.

How do you think about, Madhav?

venumusham:
- Sure, Will push a rebase commit

- For loading local file residing in browser process (device-A) from network process on device B.
  We might have to implement it by creating seperate tcp connection between browser and network process. We can submit this change in differnt CL.

venumusham:
- Rebase complete
- With this CL, network loading will be enabled by default, three processes are needed to test
1. device A: browser: 
    $ out/Default/chrome --enable-castanets www.naver.com
2. device B: renderer
    $ out/Default/chrome --enable-castanets=107.108.206.136 --type=renderer
3. device C: utility(network)
    $ out/Default/chrome --enable-castanets=107.108.206.136 --type=utility

suyambulingamrm:
Currently we will face some challenge when running network process on android.  
1. Processes are launched as activity on android.
As i checked on M69, 
    i. When we launch renderer on android, it starts a activity.
    ii. When we again launch utility on android, it kills the previous activity (renderer) and starts utility process (same behavior if we launch utility first and then renderer).
2. Also shared memory is not file based in android as posix.

PR #54, opened 2019-02-05
Implement windows port for castanets.

suyambulingamrm:
This CL is initial changes required for basic rendering in
windows port.

### DIFF
--- a/content/browser/renderer_host/media/audio_sync_reader.cc
+++ b/content/browser/renderer_host/media/audio_sync_reader.cc
@@ -10,8 +10,10 @@
 #include <utility>
 #if defined(CASTANETS)
 #include <errno.h>
+#if !defined(OS_WIN)
 #include <poll.h>
 #include <sys/socket.h>
+#endif
 #endif
 
 #include "base/command_line.h"
@@ -41,8 +43,7 @@ enum AudioGlitchResult {
 };
 
 void LogAudioGlitchResult(AudioGlitchResult result) {
-  UMA_HISTOGRAM_ENUMERATION("Media.AudioRendererAudioGlitches",
-                            result,
+  UMA_HISTOGRAM_ENUMERATION("Media.AudioRendererAudioGlitches", result,
                             AUDIO_RENDERER_AUDIO_GLITCHES_MAX + 1);
 }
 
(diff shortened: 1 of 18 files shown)


mojo/edk/system/broker_win.cc
```diff
@@ -156,5 +180,5 @@ scoped_refptr<PlatformSharedBuffer> Broker::GetSharedBuffer(size_t num_bytes) {
   return nullptr;
 }
 
-}  // namespace edk
+}  // namespace mojo
```
Sunwoo-Nam: It seems namespace mojo is closed two times. Is there no compile error?

mojo/edk/system/broker_win.cc
```diff
@@ -85,10 +103,10 @@ Channel::MessagePtr WaitForBrokerMessage(PlatformHandle platform_handle,
   }
 
   return message;
-}
-
 }  // namespace
 
+}  // namespace edk
+
```
suyambulingamrm: Empty namespace is closed here.

mojo/edk/system/broker_win.cc
suyambulingamrm: There are three namespaces in the file. 1. mojo, 2. edk. 3. a empty name space.
3 is closed above as I mentioned.
In this line edk and mojo namespaces are closed as usual.
Some how Visual studio is integrated with the clang-format. Once I close a file auto format is done.
That is the reason many files have formatting changes also.

Sunwoo-Nam:
You have already commented about automatic clang formatting.. Sorry I missed it.
I agree with your solution removing files.

mojo/edk/system/broker_win.cc
suyambulingamrm: Corrected on latest patch set.

content/browser/renderer_host/media/audio_sync_reader.cc
```diff
@@ -187,11 +190,15 @@ void AudioSyncReader::RequestMoreData(base::TimeDelta delay,
   }
 
 #if defined(CASTANETS)
-  size_t sent_bytes = HANDLE_EINTR(send(accept_handle_.get().handle, &control_signal, sizeof(control_signal), MSG_NOSIGNAL));
+  size_t sent_bytes = 0;
+#if !defined(OS_WIN)
+  HANDLE_EINTR(send(accept_handle_.get().handle, &control_signal,
```
Sunwoo-Nam: I think it is better to put the return value of send() in 'sent_bytes'.
sent_bytes = HANDLE_EINTR(send(accept_handle_.get().handle, ...

suyambulingamrm: Yes, its a mistake. Thanks for pointing.

PR #56, opened 2019-03-04
Castanets initial bringup commit.

suyambulingamrm:
1. Channel, broker connections are adapted as per M69 changes.
2. Changes related to Font Config IPC, fallback fonts are adapted to
font service.
3. Network Service is considered default for castanets 69, it is enabled
alongside renderer process. File based shared memory is used between
network service and renderer in same device.
4. Tile data is shared via bytestream.


suyambulingamrm:
Lets merge the initial commit as bringup directly, further changes can be reviewed.

venumusham:
lgtm

PR #58, opened 2019-03-14
Make tcp socket port number to be in sync with tizen environment.

sooryaR:
This CL modifies the default port number as it conflicts with
tct-manager in order to support webtct for distributed chrome.

Signed-off-by: Soorya R <soorya.r@samsung.com>

### DIFF
--- a/mojo/edk/embedder/tcp_platform_handle_utils.h
+++ b/mojo/edk/embedder/tcp_platform_handle_utils.h
@@ -17,9 +17,9 @@ namespace mojo {
 namespace edk {
 
 const size_t kCastanetsAudioSyncPort = 7000;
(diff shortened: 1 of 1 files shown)


suyambulingamrm:
+1.
Actually we changed the port number to support webtct in tizen repo. (as tct test suite's port conflicts with castanets port). https://review.tizen.org/gerrit/c/platform/framework/web/chromium-efl/+/198874/4/mojo/edk/embedder/tcp_platform_handle_utils.h
So here also we change the port number accordingly.

suyambulingamrm:
It is necessary to merge the patch as https://review.tizen.org/gerrit/c/platform/framework/web/chromium-efl/+/198874 is already landed. 
So merging the same.

PR #59, opened 2019-03-18
fixup! Castanets initial bringup commit.

suyambulingamrm:
Currently socket() call returns socket fd as 0 occasionally.
Though 0 is valid return value, the process communication
doesnot work properly. So as a workaround create & open a dummy
file so that the subsequent socket() call returns a non 0
sock fd.

### DIFF
--- a/mojo/public/cpp/platform/tcp_platform_handle_utils_posix.cc
+++ b/mojo/public/cpp/platform/tcp_platform_handle_utils_posix.cc
@@ -26,10 +26,14 @@ namespace {
 
 base::ScopedFD CreateTCPSocket(bool needs_connection, int protocol) {
   // Create the inet socket.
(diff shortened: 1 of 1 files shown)


is46-kim:
We can avoid zero fd with following step.
  if (!socket_handle) {
    int zero_fd = socket_handle;
    socket_handle = dup(zero_fd);
    close(zero_fd);
  }
This workaround is also needed in TCPServerAcceptConnection()

suyambulingamrm:
Sure, I will incorporate this change

suyambulingamrm:
Incorporated the changes. Now using dup() instead of file open.
As if (!socket_handle) include <0 cases also, I have handled 0 cases separately.

is46-kim:
LGTM

PR #63, opened 2019-03-27
[Gpu Channel] Make CreateCommandBuffer params similar to tizen env.

suyambulingamrm:
GPUCreateCommandBufferConfig struct params are different between
castanets and tizen repo, due to which IPC failure is observed
when creating commandbuffer on renderer. This change makes the struct
same as tizen repo.


hyunaranda:
lgtm

PR #64, opened 2019-04-01
Do not delete the allocated memory when SharedMemory::Close is called.

hyowon:
SharedMemory::Close() doesn't mean that the mapped memory would be deleted.
It should only close the fd.

### DIFF
--- a/base/memory/shared_memory_posix.cc
+++ b/base/memory/shared_memory_posix.cc
@@ -357,13 +357,6 @@ SharedMemoryHandle SharedMemory::TakeHandle() {
 }
 
 void SharedMemory::Close() {
-#if defined(CASTANETS)
-  if (memory_) {
-    delete [] memory_;
-    memory_ = nullptr;
-  }
-#endif
-
   if (shm_.IsValid()) {
     shm_.Close();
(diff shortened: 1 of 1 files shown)


yg48:
Need to check memory leak. 
Memory should be released in unmap. However as far as I remember unmap was not called in some cases, and it causes memory leak. Maybe memory deletion code was a kind of workaround of this issue. If there is no momory leak without this code, then I think it's ok to merge.

suyambulingamrm:
I think, this change is required.
When we fallback to software path, currently we get crash.
[19851:19863:0402/150940.559641:FATAL:raster_buffer_provider.cc(91)] Check failed: surface.
the pixels in shared memory is deleted, which seems to be not correct.
**ClientSharedBitmapManager::AllocateSharedBitmap()** deliberately does shared memory close() to save some system resources, but it expects the mapped memory to be valid.

hyowon:
Right. This pr is required for SW compositing path to work. Creating memory when map() and deleting it when unmap() seems the right way.

Sunwoo-Nam:
As my test, there is no memory leak. Also if the leak exists, we should fix it with other solutions.

PR #71, opened 2019-04-23
Multi platform support of service discovery module

nke94:
- Add osal api for process control.
- Add renderer launching impl for windows.

Signed-off-by: nke94 <nke94@samsung.com>


service_discovery_manager/Component/mmNM/tunneling/TunDrv.cpp
```diff
@@ -55,134 +55,134 @@
...
+    return -1;
+  }
+#else
+  if ((fd = open("/dev/net/tun", O_RDWR | O_NONBLOCK)) < 0) {
+    DPRINT(COMM, DEBUG_ERROR, "Cannot Open Tunneling Driver [/dev/net/tun]\n");
+    return -1;
+  }
+#endif
+
+  memset(&ifr, 0, sizeof(ifr));
```
Chikdolman, line 82: I don't see a definition for 'ifr' if WIN32 is defined.

nke94: Currently, the source code of under tunneling is not included in win32 project (Plz Look at the vs project file).
The reason why this change included  at this commit is just changing of file mode. (previously abnormally file has +x attribute. so removed "x" attribute. )
Windows implementation for tunneling is not prepared yet. for supporting windows tunneling requires totally new implementation from virtual network adapter to virtual network adapter based device I/O.
Till that implementation will be ready, plz, skip Windows tunneling supporting with this code base.

Chikdolman (approved):
Although the TunDrv cannot be used for win32, the code looks good to me for Linux based machines

PR #72, opened 2019-04-25
Add coefficient calculator

Chikdolman:
Goal: Finds coefficient values from csv list for loading time predict

CSV consists of multiple row of values and a page loading time value at the end or the row (last column).
From polynomial regression from scikit-learn, we find coefficient values.
With the coefficient values, we can predict page loading time accurately.

### DIFF
--- a/service_discovery_manager/Tools/weightFinder/getCoefficient.py
+++ b/service_discovery_manager/Tools/weightFinder/getCoefficient.py
@@ -1,0 +1,59 @@
+#
+# Tool to predict webpage loading time.
+# The edge device with shortest loading time will be selected
+# The prediction is achieved by polynomial regression
+#
+
+import numpy as np
+import sys, csv
+from sklearn.preprocessing import polynomialFeatures
+from sklearn.linear_model import LinearRegression
+
+g_poly = PolynomialFeatures(degree=2)
+g_clf = LinearRegression()
+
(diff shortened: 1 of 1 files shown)


nke94:
Not sure why this prediction is required.
It doesn't refer input data(web page) characteristic (which element page has, how many resource page has...) and device characteristic (HW spec, status,..).
In my understanding,  It just estimate page loading time by referring actually measured loading time.
So it looks that there are no difference prediction result with mean of measured loading time

nke94 (approved):
Couldn't understand a benefit of this utility fully.
But it's just independent tool. so I don't have objection.

Chikdolman:
> Not sure why this prediction is required.
> It doesn't refer input data(web page) characteristic (which element page has, how many resource page has...) and device characteristic (HW spec, status,..).
> In my understanding, It just estimate page loading time by referring actually measured loading time.
> So it looks that there are no difference prediction result with mean of measured loading time

This code actually will refer to the HW spec, status, etc.
This prediction code utilizes various numbers and is dependent on an input csv file.
In the input csv file, HW spec such as numCores, coreFreq, networkFreq, networkLat will be included.
The last column will be a page loading time for Alexa Top 100.
All HW component specifications and page loading time become a training set for the prediction.
With the training set, this code will extract coefficient(weight) value for each HW component.

PR #74, opened 2019-04-26
Fix compile warnings

hyowon:
- gpu/command_buffer/common/command_buffer.h:131:63: warning: [chromium-style]
  virtual methods with non-empty bodies shouldn't be declared inline.

- gpu/command_buffer/client/cmd_buffer_helper.cc:422:12: warning:
  moving a local object in a return statement prevents copy elision
  [-Wpessimizing-move]


Sunwoo-Nam (approved):
LGTM

PR #82, opened 2019-05-17
[SDM] Handle exception for MonitorClient::Start failure

younghajung:
This patch deletes MonitorClient related resource in case of
MonitorClient::Start failure.
Plus, closes tcp socket when connection is failed, because it can cause
fd leak.

### DIFF
--- a/service_discovery_manager/Component/mmSH/client_runner.cpp
+++ b/service_discovery_manager/Component/mmSH/client_runner.cpp
@@ -303,14 +303,18 @@ int ClientRunner::Run() {
       meta->monitor_port = info->monitor_port;
 
       meta->client = new MonitorClient(meta->id);
-      meta->message_handle = GetThreadMsgInterface(meta->id);
-      monitor_manager.AddTail(meta);
+      if (meta->client->Start(info->address, info->monitor_port)) {
+        meta->message_handle = GetThreadMsgInterface(meta->id);
+        CSTI<CbDispatcher>::getInstancePtr()->Subscribe(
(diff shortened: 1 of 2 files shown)


service_discovery_manager/Component/mmSH/client_runner.cpp
```diff
@@ -303,14 +303,18 @@ int ClientRunner::Run() {
...
-      meta->client->DataSend(monitor_packet, strlen(monitor_packet) + 1);
+      if (meta->client->Start(info->address, info->monitor_port)) {
+        meta->message_handle = GetThreadMsgInterface(meta->id);
+        CSTI<CbDispatcher>::getInstancePtr()->Subscribe(
+            MONITOR_RESPONSE_EVENT, (void*)meta->message_handle, OnMonitorClientEvent);
+        monitor_manager.AddTail(meta);
+
+        CHAR* monitor_packet = const_cast<CHAR*>("QUERY-MONITORING");
+        meta->client->DataSend(monitor_packet, strlen(monitor_packet) + 1);
+      } else {
```
Chikdolman, line 314: Should `else` exist?
Shouldn't `meta` and `meta->client` always `SAFE_DELETE()`?

younghajung: `meta` and `meta->client` will be deleted after receiving response in `OnMonitorClientEvent` in case of success.
When `MonitorClient::Start` is failed, we should delete them here.

Chikdolman:
LGTM

Chikdolman (approved):
LGTM

PR #83, opened 2019-05-17
[SDM] Fix buffer overflow in strncpy

younghajung:
Currently, the length of source string has been used as third paramter.
It the source string is longer than the destination,
it can cause the buffer overflow.
This patch uses destinations'size as the third parameter of strncpy.


Chikdolman (approved):
LGTM

PR #85, opened 2019-05-21
Propagates a type of OS of browser process to renderer process.

Sunwoo-Nam:
Renderer process needs a type of OS of browser process.
It makes renderer process can have different behaviors each target.

### DIFF
--- a/base/base_switches.cc
+++ b/base/base_switches.cc
@@ -57,6 +57,8 @@ const char kForceFieldTrials[]              = "force-fieldtrials";
 const char kNoErrorDialogs[]                = "noerrdialogs";
 
 #if defined(CASTANETS)
+// Specifies a profile of the browser process to propagate to the renderer.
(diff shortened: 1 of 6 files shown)


suyambulingamrm (commented):
Just a general clarification, currently we may have different behaviors for tizen tv, wearable, mobile etc.
Will we face a issue, that renderer should know what kind of product browser process is running ?
If so, may be we can handle that as well.

Sunwoo-Nam:
@suyambulingamrm I had considered it before.
Actually, I need only OS types for now (for multimedia codes). I think it can be classified according to device types or product types if they are necessary. I mean I thought it is not necessary for right now..

Chikdolman (approved):
Although I think there would be a better function name than "SetBrowserOSType()" as it does not receive any parameter and just adds browser_os_type to the command_line, may be we can modify the name later.
The function itself will work fine.

PR #86, opened 2019-05-22
Do not use in-memory based shared memory for discardable memory.

suyambulingamrm:


suyambulingamrm:
There are several crashes in android renderer related to discardable memory.
It looks like some page alignment calculation is going wrong in discardable memory implementation. 
It is due to the shared memory changes we added for gpu channel. 
I thought of disabling the in-memory based changes we added for discardable memory.
Please give your suggestions.

venumusham (commented):
In resource heavy pages(with high def images), SHM in-memory can cause memory overflow.
IMO, discardable memory should use shared memory on disk instead of RAM.

lgtm.

younghajung (commented):
LGTM

Chikdolman (approved):
LGTM

PR #88, opened 2019-05-30
Fix an infinite loop issue when playing media.

Sunwoo-Nam:
Browser process has to share hardware latency with renderer process
for audio streaming. In castanets case, browser process should sends
an audio buffer including the latency when changing it.
If it doesn't, renderer process has a latency as garbage value and
it can occur infinite loop.

### DIFF
--- a/media/audio/audio_device_thread.cc
+++ b/media/audio/audio_device_thread.cc
@@ -98,18 +98,29 @@ void AudioDeviceThread::ThreadMain() {
   callback_->InitializeOnAudioThread();
 
   uint32_t buffer_index = 0;
+#if defined(CASTANETS) && !defined(NETWORK_SHARED_MEMORY)
+  size_t buffer_size = callback_->shared_memory()->handle().GetSize();
+  uint8_t* buffer_data = new uint8_t[buffer_size];
+#endif
+
   while (true) {
     uint32_t pending_data = 0;
 #if defined(CASTANETS)
 #if defined(OS_WIN)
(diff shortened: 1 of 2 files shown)


media/audio/audio_device_thread.cc
```diff
@@ -98,18 +98,29 @@ void AudioDeviceThread::ThreadMain() {
...
 #if defined(CASTANETS)
 #if defined(OS_WIN)
     size_t bytes_read = 0;
 #else
+#if !defined(NETWORK_SHARED_MEMORY)
+    // Receive AudioOutputBuffer data to know delay time.
+    size_t buffer_bytes_read = HANDLE_EINTR(recv(
+         client_handle_.get().handle, buffer_data, buffer_size, MSG_WAITALL));
+    if (buffer_bytes_read != buffer_size)
+      break;
```
Chikdolman, line 117: What happens to the last part of the data? will the last part of data's buffer_bytes_read == buffer_size?
if not, doesn't it skip the last data by not copying to the shared_memory?

Chikdolman (commented):
I have a question. Please check.

media/audio/audio_device_thread.cc
Sunwoo-Nam: buffer_size is a size of shared memory. shared memory size is fixed initially.
BP sends a buffer and RP receives it with the same size.

Chikdolman (approved):
LGTM. This patch must be included in order to load Youtube page.

PR #95, opened 2019-06-18
Remove the map for tracking SharedMemoryHandles and its methods.

hyowon:
We can find a proper fd with mappings_ and holders_
in SharedMemoryTracker in case of default loading and sw compositing.
Thus we don't need to be tracking SharedMemoryHandles.


is46-kim (approved):
LGTM:)

PR #96, opened 2019-06-19
Make chromium original behavior to run properly.

hyowon:
This CL implements that the original behavior based on
Unix domain socket will run by using --enable-forking switch.

### DIFF
--- a/content/browser/child_process_launcher_helper_linux.cc
+++ b/content/browser/child_process_launcher_helper_linux.cc
@@ -26,22 +26,20 @@ namespace internal {
 
 base::Optional<mojo::NamedPlatformChannel>
 ChildProcessLauncherHelper::CreateNamedPlatformChannelOnClientThread() {
-  DCHECK_CURRENTLY_ON(client_thread_id_);
 #if defined(CASTANETS)
-  if (GetProcessType() == switches::kRendererProcess) {
-    mojo::NamedPlatformChannel::Options options;
-    options.port = mojo::kCastanetsRendererPort;
-    mojo::NamedPlatformChannel channel(options);
-    return std::move(channel);
-  }
-  else if (GetProcessType() == switches::kUtilityProcess) {
-    mojo::NamedPlatformChannel::Options options;
-    options.port = mojo::kCastanetsUtilityPort;
-    mojo::NamedPlatformChannel channel(options);
-    return std::move(channel);
-  } else
+  DCHECK_CURRENTLY_ON(client_thread_id_);
+  if (base::CommandLine::ForCurrentProcess()->HasSwitch(switches::kEnableForking))
     return base::nullopt;
+
+  mojo::NamedPlatformChannel::Options options;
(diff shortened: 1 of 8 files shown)


mojo/core/node_controller.cc
```diff
@@ -368,19 +368,17 @@ void NodeController::SendBrokerClientInvitationOnIOThread(
     // Unix Domain Socket
     PlatformChannel node_channel;
     node_connection_params = ConnectionParams(node_channel.TakeLocalEndpoint());
-    // BrokerHost owns itself.
-    BrokerHost* broker_host =
```
is46-kim: BrokerHost can communicate with Broker using by UNIX domain socket.
How about using BrokerHost instead of BrokerCastanets for IPC?

mojo/core/broker_castanets.h
```diff
@@ -89,6 +89,7 @@ class BrokerCastanets : public Channel::Delegate,
   void StartChannelOnIOThread();
 
   bool host_;
+  bool tcp_connection_ = false;
```
is46-kim, line 92: It seems simpler to use BrokerCastanets only for TCP Socket without |tcp_connection_| flag.

mojo/core/node_controller.cc
hyowon: In that case, when --enable-forking is enabled, BrokerHost will be running on the browser side while BrokerCastanets will be running on the renderer side. We expect that will be working well. I'll modify and apply that.

mojo/core/broker_castanets.h
hyowon: On the renderer side, BrokerCastanets has to cover both TCP and Unix-domain connection. The connection type check is needed.

mojo/core/node_controller.cc
hyowon: Done. One diff has been added in BrokerHost due to the BUFFER_INIT message. If any other diff occurs in BrokerHost further, we need to consider using BrokerCastanets on both processes regardless of the connection type.

content/browser/child_process_launcher_helper_linux.cc
```diff
@@ -26,22 +26,21 @@ namespace internal {
...
-    mojo::NamedPlatformChannel::Options options;
+  if (GetProcessType() == switches::kUtilityProcess)
     options.port = mojo::kCastanetsUtilityPort;
-    mojo::NamedPlatformChannel channel(options);
-    return std::move(channel);
-  } else
-    return base::nullopt;
+
+  mojo::NamedPlatformChannel channel(options);
+  return std::move(channel);
```
younghajung: I think we can just return `channel`, because the return value is rvalue.

hyowon: You're right. There is a move assignment operator for NamedPlatformChannel. That can be simply one line code. return mojo::NamedPlatformChannel(options);

is46-kim (approved):

PR #97, opened 2019-06-21
Apply the offset to memcpy() in SyncSharedBuffer().

hyowon:
This CL corrects my mistake. The loading can be completely done now.
And I've added 'const' to SyncData() in DataPipeProducerHandle.


pjh0718 (approved):
LGTM.

pjh0718 (commented):
Additionally, we could do the same type casting for 
mojo/core/broker_castanets.cc:186,
uint8_t* source = static_cast<uint8_t*>(mapped_memory->memory());
like you did in this change for consistency between them.
memcpy(extra_data, static_cast<uint8_t*>(mapping.memory()) + offset, sync_size);

And the local variable 'uint8_t* source' seems redundant.

Sunwoo-Nam (approved):
LGTM

PR #98, opened 2019-06-24
Remove non-Castanets codes from BrokerCastanets::GetWritableSharedMemoryRegion.

joonghunpark:


mojo/core/broker_castanets.cc
```diff
@@ -278,61 +278,6 @@ base::WritableSharedMemoryRegion BrokerCastanets::GetWritableSharedMemoryRegion(
       base::subtle::PlatformSharedMemoryRegion::CreateWritable(num_bytes);
   base::WritableSharedMemoryRegion r = base::WritableSharedMemoryRegion::Deserialize(std::move(region));
   return r;
```
hyowon, line 280: We have to remove this above code, or only when tcp_connection_ is true, we will do this direct allocation.

mojo/core/broker_castanets.cc
```diff
@@ -278,61 +278,6 @@ base::WritableSharedMemoryRegion BrokerCastanets::GetWritableSharedMemoryRegion(
...
-
```
hyowon, line 281: The codes below are needed for original behavior even though we are using no-sandbox...

hyowon (changes requested):
Wouldn't it be better keeping the existing code in Chromium?

joonghunpark:
According to hyowon's comment, abandon this PR for now.

PR #99, opened 2019-06-25
Return valid results in utility methods for IPC messages.

hyowon:
When invalid handles are received via Mojo connected with TCP/IP sockets, we have to allocate a shared memory directly and return a valid result outside of Mojo.

### DIFF
--- a/ipc/ipc_message_utils.cc
+++ b/ipc/ipc_message_utils.cc
@@ -46,6 +46,10 @@
 #include "mojo/public/cpp/system/scope_to_message_pipe.h"
 #endif
 
+#if defined(CASTANETS)
+#include "base/memory/shared_memory_helper.h"
+#endif
+
 namespace IPC {
 
 namespace {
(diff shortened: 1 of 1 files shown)


joonghunpark (approved):
LGTM

ipc/ipc_message_utils.cc
```diff
@@ -753,6 +757,19 @@ bool ParamTraits<base::SharedMemoryHandle>::Read(const base::Pickle* m,
   *r = base::SharedMemoryHandle(mach_port_mac.get_mach_port(),
                                 static_cast<size_t>(size), guid);
 #elif defined(OS_POSIX)
+#if defined(CASTANETS)
+  if (static_cast<internal::PlatformFileAttachment*>(attachment.get())->file()
+      == -1) {
```
Chikdolman, line 762: [Q] Is there a possibility for attachment.get() to return NULL? If there is, attachment.get()->file() might cause seg fault.

Chikdolman (commented):
Please check the comment

ipc/ipc_message_utils.cc
joonghunpark: @Chikdolman I think if at :730
if (!m->ReadAttachment(iter, &attachment))
    return false;
fails, then it would do early return.
So IMO we don't have to consider nullptr case.

@hyowon Could kCastanetsHandle defined in mojo/public/c/system/platform_handle.h be used instead of -1?

ipc/ipc_message_utils.cc
```diff
@@ -983,6 +1000,17 @@ bool ParamTraits<base::subtle::PlatformSharedMemoryRegion>::Read(
       return false;
     }
   }
+#if defined(CASTANETS)
+  if (static_cast<internal::PlatformFileAttachment*>(attachment.get())->file()
+      == -1) {
```
joonghunpark, line 1005: @hyowon Could kCastanetsHandle defined in mojo/public/c/system/platform_handle.h be used instead of -1?

ipc/ipc_message_utils.cc
hyowon: @Chikdolman No, it's not possible to access out of bound of an IPC message in this method.
@joonghunpark To add #include mojo/public/c/system/platform_handle.h is needed for that. I think kCastanetsHandle would be better to move to shared_memory_helper.h or some other file. I have no idea where is best. Why don't we just open a new PR for that?

ipc/ipc_message_utils.cc
hyowon: https://github.com/Samsung/Castanets/pull/99#discussion_r297466285

Chikdolman (approved):
LGTM

PR #101, opened 2019-06-27
Add BUFFER_SYNC_ACK.

hyowon:
- Sync caller waits until receiving BUFFER_SYNC_ACK.
- Right after data copy is complete in OnBufferSync(), send BUFFER_SYN_ACK.
- [Workaround] We must not call wait in the IO thread. In case of the browser-side loader, however, Sync() is called in the IO thread. The workaround code using ThreadChecker is to avoid it.
- [Todo] Call WaitSync() when trying to read a shared memory. That's performance could be better than SYNC_ACK approach.

### DIFF
--- a/mojo/core/broker_castanets.cc
+++ b/mojo/core/broker_castanets.cc
@@ -6,10 +6,6 @@
 
 #include <fcntl.h>
 #include <sys/mman.h>
-#include <unistd.h>
-
-#include <utility>
-#include <vector>
 
 #include "base/bind.h"
 #include "base/logging.h"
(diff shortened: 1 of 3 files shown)


mojo/core/broker_castanets.cc
```diff
@@ -206,27 +183,39 @@ bool BrokerCastanets::SyncSharedBuffer(
...
 
-  VLOG(2) << "Send Sync" << mapping.guid() << " offset: " << offset
+  VLOG(2) << "Send Sync" << guid << " offset: " << offset
           << ", sync_size: " << sync_size
           << ", buffer_size: " << buffer_sync->buffer_bytes;
   channel_->Write(std::move(out_message));
-  return true;
+
+  if (sync_wait)
+    sync_wait->Wait();
```
is46-kim, line 218: Is it possible to receive ACK before Wait() call?

hyowon: No. We immediately call Wait() after Send().

is46-kim (approved):

Sunwoo-Nam (approved):
LGTM!

mojo/core/broker_castanets.cc
```diff
@@ -262,6 +251,13 @@ void BrokerCastanets::OnBufferSync(uint64_t guid_high, uint64_t guid_low,
   memcpy(ptr + offset, data, sync_bytes);
   munmap(ptr, sync_bytes + offset);
 
+  BufferSyncAckData* sync_ack;
+  Channel::MessagePtr out_message = CreateBrokerMessage(
+      BrokerMessageType::BUFFER_SYNC_ACK, 0, 0, &sync_ack);
+  sync_ack->guid_high = guid_high;
+  sync_ack->guid_low = guid_low;
+  channel_->Write(std::move(out_message));
+
```
Chikdolman, line 260: Is it possible that BUFFER_SYNC_ACK gets skipped by line 234 even if a BUFFER_SYNC_ACK msg must be sent?

Chikdolman (commented):
Please check the question

PR #135, opened 2019-07-02
Remove workaround code for Castanets specific process.

hyowon:
- Use kCastanetsProcessHandle instead of the hard-coded process handle.
- Add process_castanets.cc to handle IsValid() checks and
  to avoid system API calls.
- Remove some code branches by defined(CASTANETS).

### DIFF
--- a/base/process/kill_posix.cc
+++ b/base/process/kill_posix.cc
@@ -28,11 +28,11 @@ TerminationStatus GetTerminationStatusImpl(ProcessHandle handle,
                                            bool can_block,
                                            int* exit_code) {
   DCHECK(exit_code);
-
 #if defined(CASTANETS)
-  if (exit_code)
+  if (handle == kCastanetsProcessHandle) {
     *exit_code = 0;
-  return TERMINATION_STATUS_STILL_RUNNING;
+    return TERMINATION_STATUS_NORMAL_TERMINATION;
+  }
 #endif
   int status = 0;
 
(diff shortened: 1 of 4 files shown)


base/process/process_castanets.cc
```diff
@@ -0,0 +1,394 @@
+// Copyright 2019 Samsung electronics. All rights reserved.
```
younghajung: It seems that this source file is forked from process_posix.cc, why didn't you add your change to process_posix.cc?
My concern is that we will need additional source files for another OS types as well, such as process_castanets_win.cc.

hyowon: About five methods in process_posix.cc require changes by Castanets. As you mentioned, process_castanets_win.cc may be required too. I'm not sure which is a better way - to add #if codes into existing files or to add new *_castanets_* files.

base/process/process_castanets.cc
```diff
@@ -0,0 +1,394 @@
...
+  bool exited = WaitForExitWithTimeoutImpl(Handle(), &local_exit_code, timeout);
+  if (exited) {
+    Exited(local_exit_code);
+    if (exit_code)
+      *exit_code = local_exit_code;
+  }
+  return exited;
+}
+
+void Process::Exited(int exit_code) const {}
```
Chikdolman: What's the need of this function? Is this implemented because of pure virtual method?

base/process/process_castanets.cc
Chikdolman: @younghajung @hyowon What could be the advantage/disadvantage of adding a new file for each platform?

Chikdolman (commented):
A bit of discussion needed. Others look ok to me.

base/process/process_castanets.cc
hyowon: If there are too many Castanets specific code in a class(file), I think creating a separate file would be slightly better. In this case, it is not quite necessary. I will update PR without adding a new file.

base/process/process_castanets.cc
hyowon: Every method in the Process class needs an implementation.

younghajung (approved):
LGTM

Chikdolman (approved):
LGTM

PR #138, opened 2019-07-04
Support communication between linux and android

suyambulingamrm:
i. Cherry picked c26cc556 from castanets_63.
ii. MOJO_CORE_LEGACY_PROTOCOL changes BrokerMessageHeader, which makes browser process's broker msgs left unidentified in android.
iii. Disable scrambling msg ids for linix as well.

### DIFF
--- a/ipc/ipc_message_macros.h
+++ b/ipc/ipc_message_macros.h
@@ -256,7 +256,7 @@
   struct IPC_MESSAGE_EXPORT msg_name##_Meta {                       \
     using InTuple = in_tuple;                                       \
     using OutTuple = out_tuple;                                     \
-    enum { ID = IPC_MESSAGE_ID() };                                 \
+    enum { ID = IPC_MESSAGE_ID(#msg_name) };                        \
     static const IPC::MessageKind kKind = IPC::MessageKind::kind;   \
     static const char kName[];                                      \
   };                                                                \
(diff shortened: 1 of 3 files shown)


ipc/ipc_message_macros.h
```diff
@@ -314,9 +314,38 @@
 // Note: we currently use __LINE__ to give unique IDs to messages within
 // a file.  They're globally unique since each file defines its own
 // IPC_MESSAGE_START.
-#define IPC_MESSAGE_ID() ((IPC_MESSAGE_START << 16) + __LINE__)
+#if defined(CASTANETS)
+// Use HASH_MSG_NAME() instead of __LINE__ for unique IDs.
+// The HASH_MSG_NAME() function should ensure unique IDs without collisions.
+// TODO: If hash IDs of messages have collisions, a build error will occur.
+constexpr uint32_t HASH_MSG_NAME(const char* s) {
```
hyowon, line 321: How about using /base code?
uint32_t PersistentHash(const std::string&); in base/hash.h

suyambulingamrm: It seems to be expecting a function marked with constexpr.
"note: non-constexpr function PersistentHash cannot be used in a constant
expression."
Will it be OK to change its definition and use ?

hyowon: I didn't notice constexpr. Compile-time string hash looks better than PersistentHash call. Let's leave it without any change.

hyowon (approved):

is46-kim (approved):

PR #139, opened 2019-07-08
This change fixes the compile error of non-Castanets build

joonghunpark:
and Castanest debug build(with the flag, enable_nacl=true).

Below changes are included.

1) Guard Castanets specific functions in mojo entrypoints' g_thunks.
2) Remove debug build compile error messages by guarding the corresponding codes.

### DIFF
--- a/content/browser/service_manager/service_manager_context.cc
+++ b/content/browser/service_manager/service_manager_context.cc
@@ -673,7 +673,6 @@ ServiceManagerContext::ServiceManagerContext(
         base::BindRepeating(&base::ASCIIToUTF16, "Visuals Service");
   }
 
-#if !defined(CASTANETS)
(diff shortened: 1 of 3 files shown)


hyowon (changes requested):
If enable_castanets = true and enable_nacl = true, how about printing a warning message and forcing enable_nacl = false? This is because for now, we do not plan to support NaCl in castanets mode.

And this PR seems to have some commits from other issues. It would be nice to separate the branch and PR.

joonghunpark:
Ok, I will separate the branch and revise this change only considering debug build with enable_nacl = false. About warning for both of enable_castanets = true and enable_nacl = true, we could handle it as a separate commit.

joonghunpark:
I re-uploaded the change to address debug build error messages only. Could you look at this again?

PR #148, opened 2019-07-18
Fix the issue that broken tiles are shown.

Sunwoo-Nam:
This issue occurs when resources are reused for tiles of different sizes.
If a mapped shared memory size is bigger than a tile size,
memory sync should be done with full shared memory size.
Memory sync is not supporting partial sync.

### DIFF
--- a/cc/raster/one_copy_raster_buffer_provider.cc
+++ b/cc/raster/one_copy_raster_buffer_provider.cc
@@ -364,7 +364,7 @@ void OneCopyRasterBufferProvider::PlaybackToStagingBuffer(
         base::CommandLine::ForCurrentProcess()->GetSwitchValueASCII("type")) {
       mojo::SyncSharedMemoryHandle(
           buffer->GetHandle().handle.GetGUID(), 0,
-          raster_full_rect.width() * raster_full_rect.height() * 4);
+          staging_buffer->size.width() * staging_buffer->size.height() * 4);
     }
 #endif
     buffer->Unmap();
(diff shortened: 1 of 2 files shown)


yg48 (commented):
LGTM :)

hyowon (approved):
For now, when rasterizing and syncing a tile, we have to send all bytes of the shared memory used by the tile. An additional improvement task: https://github.com/Samsung/Castanets/issues/150

Chikdolman (approved):
We now have two reviews. I believe YG made a mistake to comment instead of approving.

suyambulingamrm:
Can similar issue occur for downloaded resource transfer as well ?

hyowon:
@suyambulingamrm
> Can similar issue occur for downloaded resource transfer as well?

You can see that the offset is 0 and the sync_size is the total size of the shared memory when calling the SyncSharedMemoryHandle() method in tile_manager.cc and one_copy_raster_buffer_provider.cc. On the other hand, the resource loader performs synchronization as much as the byte size written to the data_pipe. (See DataPipeProducerDispatcher::SyncData() in data_pipe_producer_dispatcher.cc.) So to summarize, the resource loader is already doing a partial synchronization.

hyowon:
@suyambulingamrm
> Can similar issue occur for downloaded resource transfer as well?

Currently, there is a problem that some resources, especially such as images, are not displayed properly. Have you ever checked that those resources are downloaded (or cached?) resources?

PR #152, opened 2019-07-23
Add CastanetsMemoryMapping class.

hyowon:
- replace the simple struct MappingInfo.
- keep all memory pointers by multiple map().
- has map/unmap methods for sync.

### DIFF
--- a/base/memory/castanets_memory_mapping.h
+++ b/base/memory/castanets_memory_mapping.h
@@ -7,8 +7,8 @@
 
 #include <vector>
 
-#include "base/unguessable_token.h"
 #include "base/memory/ref_counted.h"
+#include "base/unguessable_token.h"
 
 namespace base {
 
@@ -16,7 +16,8 @@ class BASE_EXPORT CastanetsMemoryMapping
     : public RefCountedThreadSafe<CastanetsMemoryMapping> {
  public:
   static scoped_refptr<CastanetsMemoryMapping> Create(
-      const UnguessableToken& id, size_t size);
+      const UnguessableToken& id,
+      size_t size);
 
   void AddMapping(void* address);
   void RemoveMapping(void* address);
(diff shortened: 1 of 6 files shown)


base/memory/shared_memory_tracker.cc
```diff
@@ -117,16 +103,33 @@ void SharedMemoryTracker::DecrementMemoryUsage(
...
+void SharedMemoryTracker::AddMapping(
+    const UnguessableToken& guid, size_t size, void* ptr) {
+  auto it = mappings_.find(guid);
+  if (it == mappings_.end()) {
+    scoped_refptr<CastanetsMemoryMapping> castanets_mapping =
+        CastanetsMemoryMapping::Create(guid, size);
+    castanets_mapping->AddMapping(ptr);
+    mappings_.emplace(guid, castanets_mapping);
+  } else
+    it->second->AddMapping(ptr);
```
pjh0718, line 119: In case of chromium coding style, one of if else phrase has its parentheses, the others have their own parentheses also.

'git cl format' command will do that for you I guess.

hyowon: I've reformatted this commit by using git cl format.

is46-kim, line 119: |size| is always the same for the same guid?
How about adding CHECK() to compare |size|?

hyowon: Good point. I've added check code.

is46-kim (approved):

base/memory/castanets_memory_mapping.h
```diff
@@ -0,0 +1,47 @@
...
+ private:
+  friend class base::RefCountedThreadSafe<CastanetsMemoryMapping>;
+
+  CastanetsMemoryMapping(const UnguessableToken& id, size_t size);
+  ~CastanetsMemoryMapping();
+
+  UnguessableToken guid_;
+  size_t mapped_size_ = 0;
+
+  std::vector<void*> addresses_;
```
younghajung, line 43: I think each instance of CastanetsMemoryMapping is unique information, so how about making it uncopyable and unassignable? We can do it using DISALLOW_COPY_AND_ASSIGN macro.

hyowon: Good point. I've added the macro.

younghajung (approved):
LGTM

PR #153, opened 2019-07-23
Enable WebGL and Canvas2D on Castanets.

Sunwoo-Nam:
To enable WebGL, shared memories of the result, bucket and some data
need to be synchronized between GPU process and remote renderer process.
FYI, I tried to run some webgl sites (Aquarium, Toonshading) for test and
the performance was up to 30fps.

### DIFF
--- a/gpu/command_buffer/client/implementation_base.cc
+++ b/gpu/command_buffer/client/implementation_base.cc
@@ -16,6 +16,11 @@
 #include "gpu/command_buffer/client/shared_memory_limits.h"
 #include "gpu/command_buffer/common/sync_token.h"
 
+#if defined(CASTANETS)
+#include "base/command_line.h"
+#include "mojo/public/cpp/system/platform_handle.h"
+#endif
+
 namespace gpu {
 
 #if !defined(_MSC_VER)
@@ -173,7 +178,27 @@ gpu::ContextResult ImplementationBase::Initialize(
 
 void ImplementationBase::WaitForCmd() {
   TRACE_EVENT0("gpu", "ImplementationBase::WaitForCmd");
+#if defined(CASTANETS)
+  // Synchronize the result share memory because the result may have been
+  // preset.
+  if (std::string("renderer") ==
+      base::CommandLine::ForCurrentProcess()->GetSwitchValueASCII("type")) {
+    mojo::SyncSharedMemoryHandle(transfer_buffer_->shared_memory_guid(),
+                                 GetResultShmOffset(), kMaxSizeOfSimpleResult);
+  }
+#endif
+
   helper_->Finish();
+
+#if defined(CASTANETS)
+  // Request to synchronize shared memory of transfer buffer to get the result
+  // of api.
+  if (std::string("renderer") ==
+      base::CommandLine::ForCurrentProcess()->GetSwitchValueASCII("type")) {
+    helper_->CommandBufferHelper::RequestSyncTransferBuffer(
+        GetResultShmId(), GetResultShmOffset(), kMaxSizeOfSimpleResult);
+  }
+#endif
 }
 
 void* ImplementationBase::GetResultBuffer() {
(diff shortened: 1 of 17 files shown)


gpu/command_buffer/client/implementation_base.cc
```diff
@@ -207,6 +232,14 @@ bool ImplementationBase::GetBucketContents(uint32_t bucket_id,
                           buffer.size(), buffer.shm_id(), buffer.offset());
   WaitForCmd();
   uint32_t size = *result;
+#if defined(CASTANETS)
+  // Request to synchronize shared memory of transfer buffer to get bucket data.
+  if (std::string("renderer") ==
+      base::CommandLine::ForCurrentProcess()->GetSwitchValueASCII("type")) {
```
yg48, line 238: How about adding function for this? 
i.e.> IsRendererProcess(), IsBrowserProcess() ...

Sunwoo-Nam: It's a great idea. I'll try to make another patch for it because this function has been applied to other patches before.

yg48: OK :)

Sunwoo-Nam:
Enable Canvas2D patch was committed additionally.

gpu/command_buffer/client/implementation_base.cc
suyambulingamrm: Just a clarification, RequestSyncTransferBuffer() is needed both here and WaitForCmd() as well ?

gpu/command_buffer/client/mapped_memory.h
```diff
@@ -88,6 +93,12 @@ class GPU_EXPORT MemoryChunk {
   //   token: the token value to wait for before re-using the memory.
   void FreePendingToken(void* pointer, unsigned int token) {
     allocator_.FreePendingToken(pointer, token);
+#if defined(CASTANETS)
+    if (std::string("renderer") ==
+      base::CommandLine::ForCurrentProcess()->GetSwitchValueASCII("type")) {
+      mojo::SyncSharedMemoryHandle(shm_->backing()->GetGUID(), 0, GetLargestFreeSizeWithWaiting());
```
hyowon: Check offset size.

hyowon (changes requested):

gpu/command_buffer/client/mapped_memory.h
Sunwoo-Nam: Done.

gpu/command_buffer/client/implementation_base.cc
hyowon: > i.e.> IsRendererProcess(), IsBrowserProcess() ...

It can be a good way for code cleanup. But I think it is necessary to make it work properly without checking the process type anywhere.

gpu/command_buffer/client/mapped_memory.h
hyowon: Does this always sync data from the beginning of the buffer?

Sunwoo-Nam: Sorry, I modificated this line checking the offset, but it seems it was not applied. I'll retry it.

hyowon (approved):
Adding a new IPC message like GpuChannelMsg_RequestSyncTransferBuffer to control sync of data is not a proper approach. The introduction of a unified sync interface is to avoid adding this kind of code. It also causes repeated synchronization delay, which is not good for performance.
​
However, 2D/3D canvas works well now with this PR. I believe that subsequent refactoring & performance tasks will resolve these problems.​

yg48 (approved):

yg48 (approved):

gpu/command_buffer/client/implementation_base.cc
Sunwoo-Nam: @suyambulingamrm Sorry, I missed your question. Sync of WaitForCmd is for the result of commands and here is for bucket data. So both of them are needed.
But this request sync message was removed and changed on latest PR.
https://github.com/Samsung/Castanets/pull/162

suyambulingamrm: Thanks for the explanation @Sunwoo-Nam.. I got to know later :)

PR #155, opened 2019-07-24
Enable GpuCompositing on android.

suyambulingamrm:
1. GpuMemoryBufferHandle: Do not use AHardwareBufferHandle for castanets.
2. Ignore GpuBlackList, this causes serialization error as the list
   quantity is different between linux, android.
3. Unify datatypes for SurfaceHandle in linux, android.


hyowon (approved):
Android-specific code seems to be simpler and clearer than 63. Thanks. LGTM.

Sunwoo-Nam (approved):
LGTM

PR #157, opened 2019-07-25
Integration Service Discovery Manager

younghajung:
- Launch renderer process via SDM on Linux based platform
- SDM supports self detection mode for testing

### DIFF
--- a/content/browser/renderer_host/render_process_host_impl.cc
+++ b/content/browser/renderer_host/render_process_host_impl.cc
@@ -2769,8 +2769,17 @@ void RenderProcessHostImpl::AppendRendererCommandLine(
   command_line->AppendSwitchASCII(
       service_manager::switches::kServiceRequestChannelToken,
       child_connection_->service_token());
+  // TODO: This guard should be remove after resolving workaround at
+  // ChildThreadImpl::Init.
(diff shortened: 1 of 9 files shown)


content/browser/renderer_host/render_process_host_impl.cc
```diff
@@ -2769,8 +2769,12 @@ void RenderProcessHostImpl::AppendRendererCommandLine(
   command_line->AppendSwitchASCII(
       service_manager::switches::kServiceRequestChannelToken,
       child_connection_->service_token());
+  // TODO: This guard should be remove after resolving workaround at
+  // ChildThreadImpl::Init.
+#if !defined(CASTANETS)
```
hyowon: This change means that the renderer's command line does not always include the kRendererClientId switch. However, if kRendererClientId does not exist, Castanets will not work with the normal path. What about the kEnableForking check instead of code disabling with !defined(CASTANETS)?

younghajung: You're right, we should keep the normal path. Please check the latest code.

hyowon (approved):
LGTM.

Sunwoo-Nam (approved):
LGTM

PR #164, opened 2019-07-29
Set |TCP_NODELAY| option to remove delay while sending message.

Sunwoo-Nam:
Sometimes message delivery between browser process and renderer process takes
over 30ms and this makes performance degradation. The reason of this symptom is
"Nagle's algorithm" which is a means of improving the efficiency of TCP/IP.
Turn off "Nagle's algorithm" by giving |TCP_NODELAY| option.

Cherry-picked yg's commit from Tizen branch.

### DIFF
--- a/mojo/public/cpp/platform/tcp_platform_handle_utils_posix.cc
+++ b/mojo/public/cpp/platform/tcp_platform_handle_utils_posix.cc
@@ -5,6 +5,7 @@
 #include "tcp_platform_handle_utils.h"
 
 #include <arpa/inet.h>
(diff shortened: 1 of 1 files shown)


Sunwoo-Nam:
I described the issue about this PR at https://github.com/Samsung/Castanets/issues/165

yg48 (approved):
This patch will reduce network delay which is performance bottleneck of command buffer. 
It will be effective on TM1 also.

younghajung (approved):
LGTM

PR #168, opened 2019-07-30
This patch test child process launching behavior

joonghunpark:
when creating ChildProcessLauncher again and assign it as a member of
RenderProcessHostImpl.

This change is just for reference and test purpose.
I checked that with this patch,
the messages until |RenderFrameHostImpl::CommitNavigation| were queued,
and when RenderProcess was recreated with this patch,
the channel having that messages was destroyed.
So the newly created RenderProcess is connected with RenderProcessHostImpl well,
but the messages related with the navigation stuff was gone,
it leads to the situation where the new Renderer Process does nothing
because it gets no messages from the Browser Process.

We have to think again how to make connection after some timeout
between renderer and browser.


joonghunpark:
Abandoned.

PR #169, opened 2019-07-30
Use ashmem for shared memory for android.

suyambulingamrm:

### DIFF
--- a/base/memory/shared_memory_helper.cc
+++ b/base/memory/shared_memory_helper.cc
@@ -15,6 +15,10 @@
 
 #if defined(CASTANETS)
 #include "base/memory/shared_memory_tracker.h"
+#if defined(OS_ANDROID)
+#include <sys/mman.h>
+#include "third_party/ashmem/ashmem.h"
+#endif
 #endif // defined(CASTANETS)
 
 namespace base {
@@ -32,7 +36,7 @@ struct ScopedPathUnlinkerTraits {
 using ScopedPathUnlinker =
     ScopedGeneric<const FilePath*, ScopedPathUnlinkerTraits>;
 
-#if !defined(OS_ANDROID) || defined(CASTANETS)
+#if !defined(OS_ANDROID)
 bool CreateAnonymousSharedMemory(const SharedMemoryCreateOptions& options,
                                  ScopedFD* fd,
                                  ScopedFD* readonly_fd,
@@ -157,6 +161,26 @@ bool PrepareMapFile(ScopedFD fd,
   return true;
 }
 
+#elif defined(ANDROID) && defined(CASTANETS)
+bool CreateAnonymousSharedMemory(const SharedMemoryCreateOptions& options,
+                                 ScopedFD* fd,
+                                 ScopedFD* readonly_fd,
+                                 FilePath* path) {
+  // "name" is just a label in ashmem. It is visible in /proc/pid/maps.
+  fd->reset(ashmem_create_region(
+      options.name_deprecated ? options.name_deprecated->c_str() : "",
+      options.size));
+
+  int flags = PROT_READ | PROT_WRITE | (options.executable ? PROT_EXEC : 0);
+  int err = ashmem_set_prot_region(fd->get(), flags);
+  if (err < 0) {
+    DLOG(ERROR) << "Error " << err << " when setting protection of ashmem";
+    return false;
+  }
+
+  readonly_fd->reset(dup(fd->get()));
+  return true;
+}
 #endif  // !defined(OS_ANDROID)
 
 #if defined(CASTANETS)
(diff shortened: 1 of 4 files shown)


suyambulingamrm:
This PR is to discuss using ashmem for android platform.
I thought of using ashmem() apis in shared_memory_posix, platform_shared_memory_region_posix itself.
1. If we use the original files (shared_memory_android, platform_shared_memory_region_android), we face a lot of conflict because of PlatformHandle object difference in mojo side.
2. Because if 1, I feel we need to add more handlings in the code to make the interaction between linux and android.

Please give your opinions.

hyowon (commented):
First of all, I agree with your approach. The shared memory types between Android and POSIX are different:
```
#elif defined(OS_ANDROID)
  using PlatformHandle = int;
  using ScopedPlatformHandle = ScopedFD;
#else // POSIX
  using PlatformHandle = FDPair;
  using ScopedPlatformHandle = ScopedFDPair;
```

This will cause annoying code diffs. Therefore, I agree to add the ashmem related codes based on the POSIX files.

And if the browser creates a shared memory and then passes the handle to the renderer on android, we will still be opening the file in non-tmpfs without using ashmem() in CreateAnonymousSharedMemoryIfNeeded(). Could you also check this?

younghajung:
I agree too, but read-only handle will not be supported on Android, is it OK?

suyambulingamrm:
> First of all, I agree with your approach. The shared memory types between Android and POSIX are different:
> 
> ```
> #elif defined(OS_ANDROID)
>   using PlatformHandle = int;
>   using ScopedPlatformHandle = ScopedFD;
> #else // POSIX
>   using PlatformHandle = FDPair;
>   using ScopedPlatformHandle = ScopedFDPair;
> ```
> 
> This will cause annoying code diffs. Therefore, I agree to add the ashmem related codes based on the POSIX files.
> 
> And if the browser creates a shared memory and then passes the handle to the renderer on android, we will still be opening the file in non-tmpfs without using ashmem() in CreateAnonymousSharedMemoryIfNeeded(). Could you also check this?

You are right. I missed that. I will check on it.

suyambulingamrm:
> I agree too, but read-only handle will not be supported on Android, is it OK?

It seem there is no separate handle for read-only case. Chromium is using ashmem_set_prot_region() api for changing the mode of the shared memory region whenever  required. We may have to further see how to handle this.

suyambulingamrm:
Handled CreateAnonymousSharedMemoryIfNeeded() in latest patch set.

base/memory/shared_memory_helper.cc
```diff
@@ -41,6 +45,22 @@ bool CreateAnonymousSharedMemory(const SharedMemoryCreateOptions& options,
   // It doesn't make sense to have a open-existing private piece of shmem
   DCHECK(!options.open_existing_deprecated);
 #endif  // defined(OS_LINUX)
+#if defined(ANDROID) && defined(CASTANETS)
```
hyowon: Let's clean up the complex define branches.
Current:

```
#if !defined(OS_ANDROID) || defined(CASTANETS)
bool CreateAnonymousSharedMemory(...) {
  ...
#if defined(ANDROID) && defined(CASTANETS)
  // android specific impl.
#endif
  ...
}

bool PrepareMapFile(...) {
  // original impl.
}
#endif // !defined(OS_ANDROID) || defined(CASTANETS)
```

Below is my suggestion.

```
#if !defined(OS_ANDROID)
bool CreateAnonymousSharedMemory(...) {
  // original impl.
}

bool PrepareMapFile(...) {
  // original impl.
}
#elif defined(ANDROID) && defined(CASTANETS)
bool CreateAnonymousSharedMemory(...) {
  // only android specifc impl here.
}
#endif // !defined(OS_ANDROID)
```

base/memory/shared_memory_posix.cc
```diff
@@ -90,11 +90,37 @@ bool SharedMemory::CreateAndMapAnonymous(size_t size) {
...
+  }
+
+  readonly_shm_ =
+      SharedMemoryHandle(FileDescriptor(dup(fd), false),
+                         options.size, shm_.GetGUID());
+  requested_size_ = options.size;
+
+  return true;
+#else
+  if (options.size == 0) return false;
```
hyowon, line 122: Why did you move this code from the original line?

hyowon (changes requested):

base/memory/shared_memory_posix.cc
suyambulingamrm: In shared_memory_android.cc, there is no validation for size == 0 case, so I moved this line below.
And it is also mentioned as below on line 296 on shared_memory_posix.cc, so I thought size = 0 while creating ashmem is valid. 

// On Android, Map can be called with a size and offset of zero to use the
// ashmem-determined size.

hyowon (approved):
LGTM

PR #172, opened 2019-08-02
Enable Audio feature.

Sunwoo-Nam:
1. Connect with TCP socket instead of SyncSocket.
2. Sync a shared memory for audio data from RP to BP.

### DIFF
--- a/content/renderer/media/audio/mojo_audio_output_ipc.cc
+++ b/content/renderer/media/audio/mojo_audio_output_ipc.cc
@@ -11,6 +11,10 @@
 #include "mojo/public/cpp/bindings/callback_helpers.h"
 #include "mojo/public/cpp/system/platform_handle.h"
 
+#if defined(CASTANETS)
+#include "mojo/public/cpp/platform/tcp_platform_handle_utils.h"
+#endif
+
 namespace content {
 
 namespace {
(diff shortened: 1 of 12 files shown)


Sunwoo-Nam:
I'm refactoring this PR. Please do not review until update.

Sunwoo-Nam:
Now, the renderer process creates a TCP server and propagates the ip address and port number to the browser process for TCP connection.

mojo/public/cpp/platform/tcp_platform_handle_utils_posix.cc
```diff
@@ -168,4 +169,23 @@ bool TCPServerAcceptConnection(const base::PlatformFile server_socket,
 #endif  // defined(OS_NACL)
 }
 
+bool GetSelfIPAddress(std::string* address) {
+  struct hostent* host_entry;
+  char hostname[80];
+  gethostname(hostname, sizeof(hostname));
+  host_entry = gethostbyname(hostname);
```
younghajung: In my local machine, `gethostbyname` just returns 127.0.1.1.
I think `gethostbyname` is based on hosts file.
Hence, if there is no information about own IP address in hosts file, we cannot get own address.
How about using `getifaddrs`?

Sunwoo-Nam: I think It's better to remove this code :)
I am going to update PR. Thank you.

content/renderer/media/audio/mojo_audio_output_ipc.cc
```diff
@@ -233,6 +237,17 @@ void MojoAudioOutputIPC::Created(
       data_pipe->shared_memory;
   DCHECK(shared_memory_region.IsValid());
 
+#if defined(CASTANETS)
+  // If socket_handle is invalid.
+  if (socket_handle < 0) {
+    shared_memory_region_ = std::move(shared_memory_region);
```
is46-kim: How about passing |shared_memory_region| with parameter to callback function without |shared_memory_region_| member?

Sunwoo-Nam: Thank you for your point.

hyowon (approved):
It would be better if the audio service does not directly control the connection of SyncSocket. But let's merge this PR as audio works well with this implementation.

PR #179, opened 2019-08-07
Send NamedSharedMemory for Renderer Side Loader

is46-kim:
Send NamedSharedMemory instead of AnonymousSharedMemory to share the handle of SharedMemory through Message Pipe connected by TCP socket on same device.

### DIFF
--- a/base/base_switches.cc
+++ b/base/base_switches.cc
@@ -46,11 +46,6 @@ const char kForceFieldTrials[]              = "force-fieldtrials";
 // Suppresses all error dialogs when present.
 const char kNoErrorDialogs[]                = "noerrdialogs";
 
-#if defined(CASTANETS)
-// Specify distributed chrome server address.
-const char kServerAddress[]                 = "server-address";
-#endif
-
 // When running certain tests that spawn child processes, this switch indicates
 // to the test framework that the current process is a child process.
 const char kTestChildProcess[]              = "test-child-process";
@@ -121,6 +116,13 @@ const char kEnableCrashReporterForTesting[] =
 // given in base/android/library_loader/anchor_functions.h, via madvise and
 // changing the library prefetch behavior.
 const char kOrderfileMemoryOptimization[] = "orderfile-memory-optimization";
+#endif
+
+#if defined(CASTANETS)
+const char kEnableForking[] = "enable-forking";
+
+// Specify distributed chrome server address.
+const char kServerAddress[] = "server-address";
 #endif
 
 }  // namespace switches
--- a/base/base_switches.h
+++ b/base/base_switches.h
@@ -22,9 +22,6 @@ extern const char kForceFieldTrials[];
 extern const char kFullMemoryCrashReport[];
 extern const char kNoErrorDialogs[];
 extern const char kProfilingFile[];
-#if defined(CASTANETS)
-extern const char kServerAddress[];
-#endif
 extern const char kTestChildProcess[];
 extern const char kTestDoNotInitializeIcu[];
 extern const char kTraceToFile[];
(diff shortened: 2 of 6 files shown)


is46-kim:
This PR requires #182 to enable NetworkService with Unix Domain socket.

mojo/core/data_pipe_consumer_dispatcher.cc
```diff
@@ -393,13 +397,25 @@ DataPipeConsumerDispatcher::Deserialize(const void* data,
...
 #if defined(CASTANETS)
   base::subtle::PlatformSharedMemoryRegion::ScopedPlatformHandle region_handle;
+  auto guid = base::UnguessableToken::Deserialize(state->buffer_guid_high,
+                                                  state->buffer_guid_low);
+  if (state->options.flags & MOJO_CREATE_DATA_PIPE_FLAG_GUID_SHM) {
+    base::ScopedFD fd;
+    base::FilePath path;
+    if (base::CreateNamedSharedMemory(guid.ToString(),
+                                      state->options.capacity_num_bytes, &fd,
+                                      nullptr, &path, true)) {
```
joonghunpark: nullptr, &path, true /* open_only */)) would be better I think.

FYI, in these days, chromium uses enum class for this purpose because
bool type can be vague in meaning, but for now above will be enough.

e.g. 
enum class FileOpenBehavior { kOpenOnly, kWritable }

mojo/core/data_pipe_consumer_dispatcher.cc
```diff
@@ -393,13 +397,25 @@ DataPipeConsumerDispatcher::Deserialize(const void* data,
 
 #if defined(CASTANETS)
   base::subtle::PlatformSharedMemoryRegion::ScopedPlatformHandle region_handle;
+  auto guid = base::UnguessableToken::Deserialize(state->buffer_guid_high,
+                                                  state->buffer_guid_low);
+  if (state->options.flags & MOJO_CREATE_DATA_PIPE_FLAG_GUID_SHM) {
+    base::ScopedFD fd;
+    base::FilePath path;
+    if (base::CreateNamedSharedMemory(guid.ToString(),
```
joonghunpark: How about naming with CreateNamedSharedMemoryOrMapIfExists?

joonghunpark (approved):

joonghunpark (approved):
LGTM.

mojo/core/data_pipe_consumer_dispatcher.cc
is46-kim: Thank you! Looks more clear. I'll add the comment with bool value.

mojo/core/data_pipe_consumer_dispatcher.cc
is46-kim: I think the function name is enough because this function creates NamedSharedMemory.
This function has the option of creating shared memory from the existing file without creating new file.

joonghunpark: I agree. CreateNamedSharedMemory is enough.

joonghunpark (approved):

base/memory/shared_memory_helper.cc
```diff
@@ -225,6 +225,79 @@ subtle::PlatformSharedMemoryRegion CreateAnonymousSharedMemoryIfNeeded(
...
+                             bool open_only) {
+  CHECK(fd);
+  FilePath file_path;
+  if (!GetShmemTempDir(false, &file_path))
+    return false;
+  file_path = file_path.Append(name);
+  if (path)
+    *path = file_path;
+
+  if (fd) {
```
Sunwoo-Nam: Is this statement valid?
fd is already checked at line 235: CHECK(fd)

is46-kim:
There are serious issues with this PR
1. It is difficult to ensure unlinking of the file created for NamedSharedMemory.
2. Processes other than Chromium can also access resources loaded by chromium through file.

PR #183, opened 2019-08-08
[meerkat] Change project name and location

younghajung:
Project name has been changed Meerkat, and it has been moved under
third_party directory.


Chikdolman (approved):
Meerkat -> Moving to third_party

Issue #170, opened 2019-07-31
Use SharedMemory for Data Pipe without data transfer


is46-kim:
Goal : Load resources directly on renderer side device without network transfer.

Network service sends Mojo DataPipe to transfer the loaded resources to renderer. The DataPipe has a ring buffer consist of shared memory. The DataPipe is delivered by URLLoaderClient::OnStartLoadingResponseBody(mojo::ScopedDataPipeConsumerHandle body) mojo interface.

For browser side navigation scenario, Although the Renderer and Utility have Unix domain socket connections on the same device by [(#112)](https://github.com/samsung/castanets/issues/112), the shared memory handle of DataPipe can not be passed from Utility to Renderer because the Message Pipe of |URLLoaderClient| mojo interface created by Browser and connected on TCP socket channel.

The Shared Memory Handle of DataPipe should be shared between utility and renderer on the same device even through bypassing TCP socket.

is46-kim:
[WIP] Use SharedMemory named GUID for Renderer Side Resource Load.
- Create NamedSharedMemory with GUID for DataPipe in NetworkService
- Try to open the NamedSharedMemory with received GUID in Renderer
https://github.com/is46-kim/Castanets/commit/b99ce495b7af41adacaa8a92f9ec708d9629366c

Loads resources properly in Renderer with WIP patch.

is46-kim:
I requested PR https://github.com/Samsung/Castanets/pull/179
This patch sends NamedSharedMemory to share DataPipeConsumerHandle for URLLoaderClient mojo interface connected with TCP.

Issue #161, opened 2019-07-25
Incomplete web pages are shown or some subresources are not displayed correctly.


hyowon:
Currently, you can see that sometimes some resources, especially sub-resources such as images, are not displayed properly or some Javascripts print weird error messages. I cannot be one hundred percent sure, but it could be caused by the issue described below.

The problem is caused that in data pipe the consumer can read before the producer completes the write. This is because the resource loader's data is being delivered asynchronously. As looking at BrokerCastanets::SyncSharedBufferImpl(), you can see that the method does not wait for BUFFER_SYNC_ACK when it is called from the IO thread. The resource_loader's data pipe receives and writes data on the IO thread, which causes sync() calls on the same thread.

hyowon:
To solve this problem, we need to find the other way to synchronously handle with BUFFER_SYNC messages. This is also related to performance -  https://github.com/Samsung/Castanets/issues/160.

hyowon:
https://github.com/Samsung/Castanets/pull/200

PR #186, opened 2019-08-12
Call sync method in data pipe producer.

hyowon:
This CL removes all the functions that make the data pipe's client call sync directly. Data written to the ring buffer of the data pipe will attempt to sync immediately in DataPipeProducerDispatcher::EndWriteData().


hyowon:
[Warning] This commit will cause most of the source code to compile.

Sunwoo-Nam:
Looks good to me.
Is this PR only for the code refactoring? If not, is this related to the below issue?
https://github.com/Samsung/Castanets/issues/161?

hyowon:
@Sunwoo-Nam This PR is related to #161 and #170. Not only browser-side loader, but all other types of loaders will also sync (via sync inside the data pipe).

is46-kim (approved):

PR #188, opened 2019-08-13
Launch renderer process via forking at timeout

joonghunpark:
This CL for Castanets to launch renderer process if launching via tcp connection is not established within timeout.

### DIFF
--- a/content/browser/child_process_launcher_helper.cc
+++ b/content/browser/child_process_launcher_helper.cc
@@ -22,6 +22,13 @@
 #include "content/browser/android/launcher_thread.h"
 #endif
 
+#if defined(CASTANETS)
+#include "base/base_switches.h"
+#include "content/browser/renderer_host/input/timeout_monitor.h"
+
+int kTcpConnectionTimeout = 5;
+#endif
+
 namespace content {
 namespace internal {
 
@@ -76,12 +83,36 @@ ChildProcessLauncherHelper::ChildProcessLauncherHelper(
       command_line_(std::move(command_line)),
       delegate_(std::move(delegate)),
       child_process_launcher_(child_process_launcher),
+#if defined(CASTANETS)
+      tcp_connected_(false),
+#endif
       terminate_on_shutdown_(terminate_on_shutdown),
       mojo_invitation_(std::move(mojo_invitation)),
-      process_error_callback_(process_error_callback) {}
+      process_error_callback_(process_error_callback) {
+#if defined(CASTANETS)
+  tcp_success_callback_ = base::BindRepeating(
+      &ChildProcessLauncherHelper::OnCastanetsRendererLaunchedViaTcp,
+      base::Unretained(this));
+  relaunch_renderer_process_monitor_timeout_.reset(new TimeoutMonitor(
+      base::Bind(&ChildProcessLauncherHelper::OnCastanetsRendererTimeout,
+                 base::Unretained(this))));
+  relaunch_renderer_process_monitor_timeout_->Start(
+      base::TimeDelta::FromSeconds(kTcpConnectionTimeout));
+#endif
+}
 
 ChildProcessLauncherHelper::~ChildProcessLauncherHelper() = default;
+#if defined(CASTANETS)
+void ChildProcessLauncherHelper::OnCastanetsRendererTimeout() {
+  success_or_timeout_event_.Signal();
+}
 
+void ChildProcessLauncherHelper::OnCastanetsRendererLaunchedViaTcp() {
+  tcp_connected_ = true;
+  success_or_timeout_event_.Signal();
+  relaunch_renderer_process_monitor_timeout_->Stop();
+}
+#endif
 void ChildProcessLauncherHelper::StartLaunchOnClientThread() {
   DCHECK_CURRENTLY_ON(client_thread_id_);
 
@@ -121,7 +152,41 @@ void ChildProcessLauncherHelper::LaunchOnLauncherThread() {
     PostLaunchOnLauncherThread(std::move(process), launch_result);
   }
 }
+#if defined(CASTANETS)
+ChildProcessLauncherHelper::Process
+ChildProcessLauncherHelper::RetrySendOutgoingInvitation(
+    base::ProcessHandle old_process,
+    const mojo::ProcessErrorCallback& error_callback) {
+  command_line_->AppendSwitchASCII(switches::kRendererClientId,
+                                   std::to_string(child_process_id_));
 
+  DCHECK(CurrentlyOnProcessLauncherTaskRunner());
+
+  mojo_named_channel_.reset();
+  mojo_channel_.emplace();
+
+  begin_launch_time_ = base::TimeTicks::Now();
+
+  std::unique_ptr<FileMappedForLaunch> files_to_register = GetFilesToMap();
+
+  int launch_result = LAUNCH_RESULT_FAILURE;
+  base::LaunchOptions options;
+
+  Process process;
+  if (BeforeLaunchOnLauncherThread(*files_to_register, &options)) {
+    process.process = base::LaunchProcess(*command_line(), options);
+    launch_result = process.process.IsValid() ? LAUNCH_RESULT_SUCCESS
+                                              : LAUNCH_RESULT_FAILURE;
+
+    AfterLaunchOnLauncherThread(process, options);
+  }
+
+  mojo::OutgoingInvitation::Retry(old_process, process.process.Handle(),
+                                  mojo_channel_->TakeLocalEndpoint());
+
+  return process;
+}
+#endif
 void ChildProcessLauncherHelper::PostLaunchOnLauncherThread(
     ChildProcessLauncherHelper::Process process,
     int launch_result) {
@@ -153,10 +218,24 @@ void ChildProcessLauncherHelper::PostLaunchOnLauncherThread(
       DCHECK(mojo_named_channel_);
       mojo::OutgoingInvitation::Send(
           std::move(invitation), process.process.Handle(),
+#if defined(CASTANETS)
+          mojo_named_channel_->TakeServerEndpoint(), process_error_callback_,
+          tcp_success_callback_);
+#else
           mojo_named_channel_->TakeServerEndpoint(), process_error_callback_);
+#endif
     }
   }
-
+#if defined(CASTANETS)
+  if (!base::CommandLine::ForCurrentProcess()->HasSwitch(
+          switches::kEnableForking)) {
+    // If --enable-forking switch exists, we don't have to wait.
+    success_or_timeout_event_.Wait();
+    if (!tcp_connected_)
+      process = RetrySendOutgoingInvitation(process.process.Handle(),
+                                            process_error_callback_);
+  }
+#endif
   BrowserThread::PostTask(
       client_thread_id_, FROM_HERE,
       base::BindOnce(&ChildProcessLauncherHelper::PostLaunchOnClientThread,
--- a/mojo/core/broker_castanets.h
+++ b/mojo/core/broker_castanets.h
@@ -44,6 +44,10 @@ class BrokerCastanets : public Channel::Delegate, public base::SyncDelegate {
     CHECK(node_channel);
     node_channel_ = node_channel;
   }
+  void ResetNodeChannel(ConnectionParams connection_params,
+                        ScopedProcessHandle process_handle);
+
+  void ResetBrokerChannel(ConnectionParams connection_params);
 
   // Returns the platform handle that should be used to establish a NodeChannel
   // to the process which is inviting us to join its network. This is the first
@@ -77,9 +81,9 @@ class BrokerCastanets : public Channel::Delegate, public base::SyncDelegate {
   bool IsHost() const { return host_; }
 
   BrokerCastanets(base::ProcessHandle client_process,
-             ConnectionParams connection_params,
-             const ProcessErrorCallback& process_error_callback,
-             CastanetsFenceManager* fence_manager);
+                  ConnectionParams connection_params,
+                  const ProcessErrorCallback& process_error_callback,
+                  CastanetsFenceManager* fence_manager);
 
   // Send |handle| to the client, to be used to establish a NodeChannel to us.
   bool SendChannel(PlatformHandle handle);
--- a/mojo/core/channel_posix.cc
+++ b/mojo/core/channel_posix.cc
@@ -115,7 +115,19 @@ class ChannelPosix : public Channel,
 
     CHECK(server_.is_valid() || socket_.is_valid());
   }
+#if defined(CASTANETS)
+  void SetSocket(ConnectionParams connection_params) override {
+    read_watcher_.reset();
+    write_watcher_.reset();
 
+    ignore_result(socket_.release());
+    server_.TakePlatformHandle().release();
+
+    socket_ = connection_params.TakeEndpoint().TakePlatformHandle().TakeFD();
+  }
+
+  void ClearOutgoingMessages() override { outgoing_messages_.clear(); }
+#endif
   void Start() override {
 #if defined(OS_MACOSX) && !defined(OS_IOS)
     auto* relay = Core::Get()->GetMachPortRelay();
--- a/mojo/core/node_controller.cc
+++ b/mojo/core/node_controller.cc
@@ -189,7 +189,12 @@ void NodeController::SendBrokerClientInvitation(
     base::ProcessHandle target_process,
     ConnectionParams connection_params,
     const std::vector<std::pair<std::string, ports::PortRef>>& attached_ports,
+#if defined(CASTANETS)
+    const ProcessErrorCallback& process_error_callback,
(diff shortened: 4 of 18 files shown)


joonghunpark:
Currently, this CL does forking at timeout(10 seconds), but can't detect when tcp connection is established. We have to cancel the timeout manager's timer to cancel forking if launching via tcp connection has been established in this CL. To do that, we have to find the place where we can ensure that tcp connection has succeeded and signal to the timeout manager to cancel forking.

hyowon (changes requested):
Making the invitation dispatcher static and avoiding deletion will cause too many side effects.

joonghunpark:
I think we could add a function which is executed when tcp launching succeeds or timeout occurs,
in which it deletes the invitation dispatcher for clean-up. Then 'static' will not be problematic I think, and for this purpose, we have to choose where is the proper location to signal child process launching success via tcp connection.

hyowon:
Each time you launch a child process, a new invitation will be generated. I wonder how you can handle invitation dispatchers properly. I will review it again with the next patch.

joonghunpark:
In class ChildConnection::IOThreadContext,
void InitializeOnIOThread(
...
    if (connector_) {
      connector_->StartService(child_identity, std::move(service),
                               std::move(pid_receiver_request));
      connector_->BindInterface(child_identity, &child_);
    }
  }

connector_->StartService above starts renderer service with process and ChildConnection is made only once per RenderProcessHostImpl.

That's why we can't make renderer process when we send OutgoingInvitation second time.

For now, I revised the CL's behavior as below.

[WIP] This CL changes child process launching behavior as below.

1) If timeout occurs waiting child process to be launched,
then uses child process launched with 127.0.0.1 connected via tcp.
2) After the first tcp connected child process,
uses --enable-forking to launch child processes to use unix domain socket.


If needed, I'd like to investigate sad tab reload function for reference.

hyowon:
See content/child/child_thread_impl.cc:511

```
if (!base::CommandLine::ForCurrentProcess()->HasSwitch(
	switches::kRendererClientId)) {
  base::CommandLine::ForCurrentProcess()->AppendSwitchASCII(
	  switches::kRendererClientId, std::to_string(1)); // workaround
  invitation = InitializeMojoIPCChannelTCP();
} else
  invitation = InitializeMojoIPCChannel();
```

hyowon:
- Fork a new child process after 3 seconds sleep.
- Retry invitation with the new process and unix domain socketpair information.
- Discard existing tcp sockets and reset to new sockets in channel_posix.

We have confirmed that the above steps work well.
Refer commit below.
https://github.com/hyowon/Castanets/commit/ed654bbbdf77d1881c1d3002f047eac437847b39

joonghunpark:
* CL Status
Currently, if launching via tcp connection is not established
within timeout(3 sec), use forked renderer process.

This CL needs to add branch not to fork if tcp connection was established.

joonghunpark:
Could you please review this change?
Now the CL works as expected.

mojo/core/node_controller.cc
```diff
@@ -508,6 +524,7 @@ void NodeController::SendBrokerClientInvitationOnIOThread(
   // called below.
 
   pending_invitations_.insert(std::make_pair(temporary_node_name, channel));
+  temp_node_name_ = temporary_node_name;
```
hyowon: We can access a node channel without this variable. See broker_castanets.h:126.

mojo/core/node_controller.cc
```diff
@@ -516,6 +533,41 @@ void NodeController::SendBrokerClientInvitationOnIOThread(
...
+    NamedPlatformChannel named_channel(options);
+    node_connection_params =
+        ConnectionParams(named_channel.TakeServerEndpoint());
+    broker_host->SendNamedChannel(named_channel.GetServerName());
+  }
+#else
+  CHECK(channel_ok);
+#endif  // defined(OS_WIN)
+
+  auto it = pending_invitations_.find(temp_node_name_);
```
hyowon: BrokerCastanets already has a node_channel pointer. See broker_castanets.h:126

mojo/core/node_controller.h
```diff
@@ -277,6 +286,7 @@ class MOJO_SYSTEM_IMPL_EXPORT NodeController : public ports::NodeDelegate,
   Core* const core_;
   const ports::NodeName name_;
   const std::unique_ptr<ports::Node> node_;
+  base::RepeatingCallback<void()> tcp_success_callback_;
```
hyowon, line 295: I think it is not appropriate for NodeController to have this callback function as a member variable. And why is it a RepeatingCallback type? Wouldn't it be better OnceCallback?

mojo/core/node_controller.h
```diff
@@ -344,6 +354,8 @@ class MOJO_SYSTEM_IMPL_EXPORT NodeController : public ports::NodeDelegate,
   // Channels to invitees during handshake.
   NodeMap pending_invitations_;
 
+  ports::NodeName temp_node_name_;
```
hyowon: Same as above comment.

content/browser/child_process_launcher_helper.cc
```diff
@@ -76,12 +79,30 @@ ChildProcessLauncherHelper::ChildProcessLauncherHelper(
...
 
 ChildProcessLauncherHelper::~ChildProcessLauncherHelper() = default;
 
+void ChildProcessLauncherHelper::OnCastanetsRendererTimeout() {
+  success_or_timeout_event_.Signal();
+}
+
+void ChildProcessLauncherHelper::OnCastanetsRendererLaunchedViaTcp() {
+  tcp_connected_ = true;
+  success_or_timeout_event_.Signal();
```
hyowon, line 112: relaunch_renderer_process_monitor_timeout_.Stop();

mojo/core/broker_castanets.h
```diff
@@ -127,6 +130,8 @@ class BrokerCastanets : public Channel::Delegate, public base::SyncDelegate {
 
   std::unique_ptr<CastanetsFenceQueue> fence_queue_;
 
+  base::OnceCallback<void()> tcp_success_callback_;
```
hyowon: Both Nodecontroller and BrokerCastanets has a callback for TCP connection success?

mojo/core/channel_posix.cc
```diff
@@ -116,6 +116,21 @@ class ChannelPosix : public Channel,
...
+
+    ignore_result(socket_.release());
+    server_.TakePlatformHandle().release();
+
+    socket_ = connection_params.TakeEndpoint().TakePlatformHandle().TakeFD();
+  }
+
+  void ClearOutgoingMessages() override {
+    base::circular_deque<MessageView> messages;
+    std::swap(outgoing_messages_, messages);
```
hyowon: Isn't it just okay to call outgoing_messages_.clear()?

hyowon (changes requested):

content/browser/child_process_launcher_helper.cc
joonghunpark: Done.

mojo/core/broker_castanets.h
joonghunpark: This is just a residue from my previous patchset.
Let me remove this.

mojo/core/channel_posix.cc
joonghunpark: Ok, I will change this as you pointed out.

mojo/core/node_controller.cc
joonghunpark: Done.

mojo/core/node_controller.cc
joonghunpark: Done. I added a getter for it.

mojo/core/node_controller.h
joonghunpark: Done.

mojo/core/node_controller.h
joonghunpark: OnceCallback::Run() may only be invoked on a non-const rvalue, i.e. std::move(callback).Run().
After std::move the member tcp_success_callback_ will be nullptr, so it could be a little odd.

I declared base::RepeatingCallback<void()> tcp_success_callback_; at node_controller
to use it in NodeController::OnAcceptInvitation.

joonghunpark (commented):
I removed temp_node_name_ and uses BrokerCastanets to retrieve node channel as you commented.
Could you please review change again?

mojo/core/broker_castanets.h
```diff
@@ -40,11 +40,15 @@ class BrokerCastanets : public Channel::Delegate, public base::SyncDelegate {
 
   ~BrokerCastanets() override;
 
+  scoped_refptr<NodeChannel> GetNodeChannel() { return node_channel_; }
+
```
hyowon: Wouldn't it be better to add ResetNodeChannel() method rather than using this getter?

mojo/core/node_controller.cc
```diff
@@ -516,6 +532,41 @@ void NodeController::SendBrokerClientInvitationOnIOThread(
...
+  }
+#else
+  CHECK(channel_ok);
+#endif  // defined(OS_WIN)
+
+  auto it = broker_hosts_.find(old_process.get());
+  if (it == broker_hosts_.end())
+    return;
+  auto channel = it->second->GetNodeChannel();
+  channel->SetSocket(std::move(node_connection_params));
```
hyowon: As I mentioned above, if we have ResetNodeChannel() method, these lines could be simpler.

mojo/core/node_controller.h
hyowon: RepeatingCallback allows calling Run() any number of time. However, this callback will have to run only once per one invitation, if not, we should think there is a bug.

```
SendInvitation...(..., OnceCallback callback) {
    connection_callback_ = std::move(callback);
}

AcceptInvitation...() {
    DCHECK(!connection_callback_.is_null());
    std::move(connection_callback_).Run();
}
```

If you still think it looks odd, let's leave it without change. This is just my suggestion.

content/browser/child_process_launcher_helper.cc
```diff
@@ -22,6 +24,8 @@
 #include "content/browser/android/launcher_thread.h"
 #endif
 
+int kTcpConnectionTimeout = 3;
```
hyowon: const and how about 5 sec as a default. My hands are too slow.

hyowon (changes requested):

mojo/core/broker_castanets.h
joonghunpark: Ok, I will add ResetNodeChannel() also.

mojo/core/node_controller.cc
joonghunpark: I will apply this in the following patchset.

content/browser/child_process_launcher_helper.cc
joonghunpark: Let me change to 5 sec.

joonghunpark: When I tried with OnceCallback, Mostly ok but sometimes the logs like below was observed.
Used LOG is,
void NodeController::OnAcceptInvitation() {
  LOG(INFO) << "tcp_success_callback_.is_null(): " << tcp_success_callback_.is_null();
}

I think the reason is some web pages including multiple cross domains uses multiple renderer process for each of these for security reason.
And for this reason, I think RepeatingCallback could be reasonable.

For your reference, I will leave the patch's url with OnceCallback impl.
https://github.com/joonghunpark/Castanets/commits/oncecallback

[6117:6154:0911/113107.155153:INFO:child_process_launcher_helper.cc(183)] PostLaunchOnLauncherThread
[6117:6117:0911/113107.155139:INFO:render_process_host_impl.cc(3914)] SKIP!!!!! RenderProcessHostImpl::CreateSharedRendererHistogramAllocator
[6117:6154:0911/113107.155253:INFO:child_process_launcher_helper.cc(213)] PostLaunchOnLauncherThread - mojo::OutgoingInvitation::Send
[6117:6120:0911/113107.156187:INFO:tcp_platform_handle_utils_posix.cc(134)] Server Sock fd for port 37149:498
[6117:6117:0911/113107.165928:INFO:render_process_host_impl.cc(3914)] SKIP!!!!! RenderProcessHostImpl::CreateSharedRendererHistogramAllocator
[6191:6191:0911/113107.618523:WARNING:chrome_main_delegate.cc(280)] process type 'renderer' should be created through the zygote.
[6191:6191:0911/113107.619544:INFO:broker_castanets.cc(147)] Connection Success: Unix Domain Socket
**[6117:6120:0911/113107.619648:INFO:node_controller.cc(971)] tcp_success_callback_.is_null(): 0**
[6192:6192:0911/113107.624037:WARNING:chrome_main_delegate.cc(280)] process type 'renderer' should be created through the zygote.
[6192:6192:0911/113107.625040:INFO:broker_castanets.cc(147)] Connection Success: Unix Domain Socket
**[6117:6120:0911/113107.625115:INFO:node_controller.cc(971)] tcp_success_callback_.is_null(): 1**

hyowon: > When I tried with OnceCallback, Mostly ok but sometimes the logs like below was observed.
> Used LOG is,
> void NodeController::OnAcceptInvitation() {
> LOG(INFO) << "tcp_success_callback_.is_null(): " << tcp_success_callback_.is_null();
> }
> 
> I think the reason is some web pages including multiple cross domains uses multiple renderer process for each of these for security reason.

Seriously? If it is true, Castanets won't work at all. Please share the test site.

> And for this reason, I think RepeatingCallback could be reasonable.

multiple renderer processes -> multiple child process launchers -> multiple invitations -> multiple callbacks BUT that doesn't mean one callback runs multiple times. As I explained, a success callback has to run only once per invitation. Do you agree with this part?

If a success callback is null in AcceptInvitation(), this may be one of the two cases below.
1) you haven't set a proper callback in SendInvitation() -> bug
2) you try to run the same callback for the second time even though the callback had been already run. (The second connection success?) -> bug

> For your reference, I will leave the patch's url with OnceCallback impl.
> https://github.com/joonghunpark/Castanets/commits/oncecallback

Let's just leave this PR with a repeating callback. If necessary, I will create a PR.

joonghunpark: >Seriously? If it is true, Castanets won't work at all. Please share the test site.

The nullptr OnceCallback occurs only on 2nd Tab. 3rd or 4th is ok it seems.
So I think my assumption about the issue was incorrect. Sorry for my immature assumption.

>multiple renderer processes -> multiple child process launchers -> multiple invitations -> >multiple callbacks BUT that doesn't mean one callback runs multiple times. As I >explained, a success callback has to run only once per invitation. Do you agree with this >part?

I agree with this.

>If a success callback is null in AcceptInvitation(), this may be one of the two cases below.

>1. you haven't set a proper callback in SendInvitation() -> bug
>2. you try to run the same callback for the second time even though the callback had >been already run. (The second connection success?) -> bug

I think this corresponds the case 2.

>Let's just leave this PR with a repeating callback. If necessary, I will create a PR.

 I see. Thank you for your review:)

joonghunpark:
I applied the comments in the previous patchset. Could you please review this change?

content/browser/child_process_launcher_helper.cc
```diff
@@ -22,6 +24,10 @@
 #include "content/browser/android/launcher_thread.h"
 #endif
 
+#if defined(CASTANETS)
+int kTcpConnectionTimeout = 5;
```
hyowon, line 29: add include here.

mojo/core/broker_castanets.h
```diff
@@ -77,9 +81,9 @@ class BrokerCastanets : public Channel::Delegate, public base::SyncDelegate {
   bool IsHost() const { return host_; }
 
   BrokerCastanets(base::ProcessHandle client_process,
-             ConnectionParams connection_params,
-             const ProcessErrorCallback& process_error_callback,
-             CastanetsFenceManager* fence_manager);
+                  ConnectionParams connection_params,
+                  const ProcessErrorCallback& process_error_callback,
+                  CastanetsFenceManager* fence_manager);
```
hyowon, line 86: I've already applied the same code cleanup. Don't put code for cleanup in this pr. One commit = one issue.

hyowon (approved):

content/browser/child_process_launcher_helper.cc
joonghunpark: Done.

mojo/core/broker_castanets.h
joonghunpark: I checked your commit below, and this change made by 'git cl format' was not applied in the commit below.
This style change was made automatically by script(although I didn't intent to do that), so I guess it's ok to go.

d5e7ff134cf743a39c7ba11618afb20c373fc104
Author: Hyowon Kim <hw1008.kim@samsung.com>
Date:   Wed Sep 11 10:07:14 2019 +0900

    Code cleanup in BrokerCastanets.

joonghunpark (commented):
All the comments left in the previous patchset has been applied.

PR #189, opened 2019-08-13
[DCHECK] ChildProcessLauncherHelper::GetFilesToMap()

Tanvir-rizvi:
For NamedPlatformChannel, channel is not created
and therefore mojo_channel_ wrapped in optional.h
is not populated and hence DCHECK.
Since the server endpoint is not connected,
PlatformChannelEndPoint is invalid. So there may be no
need to map the shared region for named channel.

Bug: 174 [GitHub.com/samsung/Catsanets]

### DIFF
--- a/content/browser/child_process_launcher_helper.cc
+++ b/content/browser/child_process_launcher_helper.cc
@@ -125,6 +125,12 @@ void ChildProcessLauncherHelper::LaunchOnLauncherThread() {
 void ChildProcessLauncherHelper::PostLaunchOnLauncherThread(
     ChildProcessLauncherHelper::Process process,
     int launch_result) {
+#if defined(CASTANETS)
+  // If mojo_named_channel_ is valid, we are trying to launch process
+  // in Castanets mode, mojo_channel_ is no longer needed.
+  if (mojo_named_channel_)
+    mojo_channel_ = base::nullopt;
+#endif
   if (mojo_channel_)
     mojo_channel_->RemoteProcessLaunchAttempted();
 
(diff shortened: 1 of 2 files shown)


content/browser/child_process_launcher_helper_posix.cc
```diff
@@ -87,11 +87,12 @@ std::unique_ptr<PosixFileDescriptorInfo> CreateDefaultPosixFilesToMap(
         base::SharedMemory::GetFdFromSharedMemoryHandle(shm));
   }
 
+#if !defined(CASTANETS)
```
hyowon: We have to ensure that the child process will work correctly even if launched in the normal path. If these codes are disabled, normal launching will fail.

hyowon (commented):
You found the exact codes needed to fix. However, I would like to suggest a simpler fix.

```
base::Optional<mojo::NamedPlatformChannel>
ChildProcessLauncherHelper::CreateNamedPlatformChannelOnClientThread() {
  if (kEnableForking) // normal launching
    return base::nullopt;
  // castanets launching
  ......
+  mojo_channel_.emplace(); // This socket pair is not used. 
+  // To avoid failure of validation check of codes afterwards.
  return mojo::namedPlatformChannel(options);  
}

void ChildProcessLauncherHelper::PostLaunchOnLauncherThread(...) {
+ #if defined(CASTANETS)
+   // if named_channel_ is valid, we are now trying to launch a process in castanets mode.
+   // mojo_channel_ is no longer needed.
+   if (mojo_named_channel_)
+     mojo_channel_ = base::nullopt;
+ #endif
  ....
}
```

Tanvir-rizvi (commented):
Updated code as per the comments. Please review. Thanks

hyowon (approved):
Thanks. LGTM.

hyowon:
Please remove the last commit. PR is automatically updated with your branch.

PR #192, opened 2019-08-20
Add null check of NavigationRequest

is46-kim:
This commit fixes a crash when loading file scheme URL with Network Service.


content/browser/frame_host/frame_tree_node.cc
```diff
@@ -414,6 +414,10 @@ bool FrameTreeNode::CommitPendingFramePolicy() {
 
 void FrameTreeNode::TransferNavigationRequestOwnership(
     RenderFrameHostImpl* render_frame_host) {
```
Sunwoo-Nam, line 416: It seems #if defined(CASTANETS) is necessary.

Sunwoo-Nam (approved):
Looks good to me.

PR #193, opened 2019-08-20
[DCHECK] AssertBlockingAllowed

Tanvir-rizvi:
base::CreateAndOpenFdForTemporaryFileInDir is called
when fd < 0.  In this scenario DCHECK is hit.
DCHECK hits when the above function is called
from a scope that disallows blocking.
To avoid this we will call the above function
using ScopedAllowIO.

Bug: Samsung#190


hyowon (approved):
LGTM

PR #195, opened 2019-08-22
Add trace logs to profile the shared memory synchronization.

Sunwoo-Nam:
The progress of the shared memory synchronization is very important
on Castanets. We want to use Chrome trace to do that and time ticks
between processes need to be synchronized.

This CL still has a problem because the network delay affects timestamps.
We better check only total memories transferred without a duration.

### DIFF
--- a/services/tracing/public/cpp/trace_event_agent.cc
+++ b/services/tracing/public/cpp/trace_event_agent.cc
@@ -22,6 +22,10 @@
 #include "services/tracing/public/cpp/tracing_features.h"
 #include "services/tracing/public/mojom/constants.mojom.h"
 
+#if defined(CASTANETS)
+#include "base/command_line.h"
+#endif
+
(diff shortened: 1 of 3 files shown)


services/tracing/public/cpp/trace_event_agent.cc
```diff
@@ -146,6 +150,13 @@ void LegacyTraceEventAgent::StartTracing(const std::string& config,
   // latency.
   base::TimeDelta time_offset = TRACE_TIME_TICKS_NOW() - coordinator_time;
   TraceLog::GetInstance()->SetTimeOffset(time_offset);
+#endif
+#if defined(CASTANETS)
+  if (std::string("renderer") ==
+      base::CommandLine::ForCurrentProcess()->GetSwitchValueASCII("type")) {
```
santhoshkerur, line 156: how about defining like this ? for more easy modifications in future.
constexpr const char kProcessTypeSwitch[] = "type";

PR #197, opened 2019-08-23
Enable to play video with MediaElement

Sunwoo-Nam:
Synchronize the gpu memory buffer in the video frame resource.
In case of a SW compositor, synchronize the SW resource.
Set the frame time to the system time of renderer process.


Sunwoo-Nam:
This PR makes the video be able to be played on Ubuntu desktop.

hyowon (approved):
LGTM

Issue #176, opened 2019-08-06
Check failed with gpu composing


is46-kim:
Check failed with gpu composing on linux. It causes a crash.
This check failure does not occur with "--disable-gpu-compositing" command line option.

< Backtrace >
[3228:3237:0806/173630.380276:FATAL:gles2_implementation.cc(1170)] Check failed: query->CheckResultsAvailable(helper_). 

#0 0x7f224b0e493c base::debug::StackTrace::StackTrace()
#1 0x7f224b045d40 logging::LogMessage::~LogMessage()
#2 0x7f223e489aa4 gpu::gles2::GLES2Implementation::GetQueryObjectValueHelper()
#3 0x7f223e49c249 gpu::gles2::GLES2Implementation::GetQueryObjectuivEXT()
#4 0x7f223e63667d cc::StagingBufferPool::AcquireStagingBuffer()
#5 0x7f223e631d09 cc::OneCopyRasterBufferProvider::PlaybackAndCopyOnWorkerThread()
#6 0x7f223e631bb1 cc::OneCopyRasterBufferProvider::RasterBufferImpl::Playback()
#7 0x7f223e675774 cc::(anonymous namespace)::RasterTaskImpl::RunOnWorkerThread()
#8 0x7f224919b6bc content::CategorizedWorkerPool::RunTaskInCategoryWithLockAcquired()
#9 0x7f224919a54a content::CategorizedWorkerPool::Run()
#10 0x7f224b0b7324 base::SimpleThread::ThreadMain()
#11 0x7f224b0f26ed base::(anonymous namespace)::ThreadFunc()
#12 0x7f2241ed36ba start_thread
#13 0x7f223fe4e41d clone

Sunwoo-Nam:
We can skip the Check() temporarily like version 63.
But after that, the below crash occurs on browser process.
It seems that BrokerCastanets is not valid.

Received signal 11 SEGV_MAPERR 00000000003f
#0 0x7f2d74f38eac base::debug::StackTrace::StackTrace()
#1 0x7f2d74f38a01 base::debug::(anonymous namespace)::StackDumpSignalHandler()
#2 0x7f2d6bb6f390 <unknown>
#3 0x7f2d69a78a10 <unknown>
#4 0x7f2d6675c1ed mojo::core::BrokerCastanets::OnBufferSync()
#5 0x7f2d6675cb42 mojo::core::BrokerCastanets::OnChannelMessage()
[...]

PR #198, opened 2019-08-23
Make SharedMemory as writable memory for Castanets

is46-kim:
SharedMemory should be created as writable memory to sync memory even if client makes SharedMemory as read only.
Writing to read-only memory causes SEGV_ACCERR crash.


Sunwoo-Nam (approved):
Looks good to me.
This PR solves the issue below.
https://github.com/Samsung/Castanets/issues/176
@hyowon I need your review.

PR #199, opened 2019-08-28
Lock between shared memory management(holder, mapping) and OnBufferSync().

hyowon:
#161 related bug fix and implementation.

PR #200, opened 2019-08-29
Fence & WaitSync

hyowon:
Castanets will introduce a new fence method to access shared memory and sync data.
blocked by https://github.com/Samsung/Castanets/pull/199


Sunwoo-Nam (approved):
Behavior is verified. LGTM.

PR #203, opened 2019-09-03
Partial synchronization of memory for partial rasterization.

Sunwoo-Nam:
On SW compositing mode, tiles are rasterized the size of the dirty area
partially. The memory, therefore, needs to be sent the partial size.
Also, enable kDisallowNonExactResourceReuse because reusing resources
should only be done if they're the same size as requested.
These will reduce memory sync usage.

### DIFF
--- a/cc/tiles/tile_manager.cc
+++ b/cc/tiles/tile_manager.cc
@@ -135,9 +135,27 @@ class RasterTaskImpl : public TileTask {
     if (std::string("renderer") ==
         base::CommandLine::ForCurrentProcess()->GetSwitchValueASCII("type")) {
       ResourcePool::SoftwareBacking* sw_backing = resource_.software_backing();
-      if (sw_backing)
-        mojo::SyncSharedMemoryHandle(sw_backing->SharedMemoryGuid(),
-            0, resource_.size().width() * resource_.size().height() * 4);
+      int bytes_per_pixel = BitsPerPixel(resource_.format()) / 8;
+      if (sw_backing &&
+          UsePartialMemorySync(content_rect_, invalid_content_rect_)) {
+        size_t offset_x =
+            (invalid_content_rect_.x() - content_rect_.x()) * bytes_per_pixel;
+        size_t offset_y =
+            (invalid_content_rect_.y() - content_rect_.y()) * bytes_per_pixel;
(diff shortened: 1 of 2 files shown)


cc/tiles/tile_manager.cc
```diff
@@ -162,6 +177,25 @@ class RasterTaskImpl : public TileTask {
   }
 
  private:
+#if defined(CASTANETS)
+  bool UsePartialMemorySync(gfx::Rect content_rect, gfx::Rect invalid_rect) {
```
hyowon, line 184: A trivial suggestion to make this function simpler.

```
  bool UsePartialMemorySync(gfx::Rect content_rect, gfx::Rect invalid_rect) {
    if (invalid_rect.IsEmpty() || content_rect == invalid_rect)
      return false;

    // If dirty area is greater than 50% of content area,
    // do not use partial sync.
    if (((float)invalid_rect.size().GetArea() / 
         (float)content_rect.size().GetArea()) > 0.5)
      return false;	  
    return true;
  }
```

cc/tiles/tile_manager.cc
```diff
@@ -135,9 +135,24 @@ class RasterTaskImpl : public TileTask {
     if (std::string("renderer") ==
         base::CommandLine::ForCurrentProcess()->GetSwitchValueASCII("type")) {
       ResourcePool::SoftwareBacking* sw_backing = resource_.software_backing();
-      if (sw_backing)
-        mojo::SyncSharedMemoryHandle(sw_backing->SharedMemoryGuid(),
-            0, resource_.size().width() * resource_.size().height() * 4);
+      if (sw_backing &&
+          UsePartialMemorySync(content_rect_, invalid_content_rect_)) {
+        size_t offset_x = (invalid_content_rect_.x() - content_rect_.x()) * 4;
```
hyowon: Instead of 4, 
int bytes_per_pixel = BitsPerPixel(resource_.format()) / 8;

cc/tiles/tile_manager.cc
Sunwoo-Nam: Thanks, done.

cc/tiles/tile_manager.cc
Sunwoo-Nam: Thanks, done.

hyowon (approved):

PR #205, opened 2019-09-06
Add sync shared memory api for 2-dimensional memory.

Sunwoo-Nam:
Until now, Broker only has the api for linear memory sync. But rasterizing
a tile partially requires synchronization of 2-dimensional memory.
For this, |width| and |stride| are added to sync data to calculate the
rectangular memory.

### DIFF
--- a/mojo/core/broker_castanets.cc
+++ b/mojo/core/broker_castanets.cc
@@ -238,6 +238,25 @@ bool BrokerCastanets::SyncSharedBuffer(
   return true;
 }
 
+bool BrokerCastanets::SyncSharedBuffer2d(const base::UnguessableToken& guid,
+                                         size_t offset,
+                                         size_t sync_size,
+                                         size_t width,
+                                         size_t stride) {
+  if (!tcp_connection_)
+    return true;
+
+  scoped_refptr<base::CastanetsMemoryMapping> mapping =
+      base::SharedMemoryTracker::GetInstance()->FindMappedMemory(guid);
+  if (!mapping)
+    return MOJO_RESULT_NOT_FOUND;
+
+  SyncSharedBufferImpl2d(guid, static_cast<uint8_t*>(mapping->GetMemory()),
+                         offset, sync_size, mapping->mapped_size(), width,
+                         stride);
+  return true;
+}
+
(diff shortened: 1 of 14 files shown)


mojo/core/broker_castanets.cc
```diff
@@ -268,6 +287,51 @@ void BrokerCastanets::SyncSharedBufferImpl(const base::UnguessableToken& guid,
...
+  size_t put_bytes = 0;
+  size_t put_offset = offset;
+
+  while (put_bytes != sync_size) {
+    memcpy(static_cast<uint8_t*>(extra_data) + put_bytes, memory + put_offset,
+           width);
+    put_offset += stride;
+    put_bytes += width;
+  }
+
```
hyowon, line 324: base::AutoLock lock(sync_lock_); is necessary here.

mojo/core/node_controller.cc
```diff
@@ -365,6 +365,18 @@ bool NodeController::SyncSharedBuffer(
...
 
+bool NodeController::SyncSharedBuffer2d(const base::UnguessableToken& guid,
+                                        size_t offset,
+                                        size_t sync_size,
+                                        size_t width,
+                                        size_t stride) {
+  if (broker_)
+    return broker_->SyncSharedBuffer2d(guid, offset, sync_size, width, stride);
+
+  // Normal path with Unix domain socket on browser(host)
```
hyowon: "broker_ is null" means that this is the browser process. The browser process is not likely to send tile(2d) data. I think we need NOTREACHED() here.

hyowon (changes requested):
This pr will help performance improvement.

mojo/core/node_controller.cc
hyowon: CHECK(broker_) also looks good.

mojo/core/broker_castanets.cc
Sunwoo-Nam: Done. Thanks.

mojo/core/node_controller.cc
Sunwoo-Nam: Done. Thanks.

hyowon (approved):

Sunwoo-Nam:
I added the fixup! commit for the crash issue.
Please review the commit either.

PR #211, opened 2019-09-11
Fix the scroll performance issue with GPU compositing.

Sunwoo-Nam:
When playback is in progress, it tries to get the query result in
|WaitForQueryResult| in staging_buffer_pool.cc. It causes delay of the raster
and the query isn't handled on Castanets now, we should skip it.


hyowon (approved):
What about the [Workaround] prefix in the commit title?

PR #227, opened 2019-09-18
Fix the issue for the invisible scrollbar.

Sunwoo-Nam:
UI resource has own shared memory which should be synchronized.
This issue occurs with SW compositing.


Sunwoo-Nam:
There is WaitSync for SharedBitmap so additional WaitSync isn't necessary.

hyowon (approved):
LGTM

PR #230, opened 2019-09-19
Allow the node channel to connect using SSL.

hyowon:
This CL adds "--secure-connection" switch to allow the node channel to connect securely by using BoringSSL. Temporarily a self-signed certificate file is being used.

### DIFF
--- a/mojo/core/channel_posix.cc
+++ b/mojo/core/channel_posix.cc
@@ -29,6 +29,7 @@
 #include "base/memory/shared_memory_tracker.h"
 #include "mojo/public/cpp/platform/secure_socket_utils_posix.h"
 #include "mojo/public/cpp/platform/tcp_platform_handle_utils.h"
+#include "third_party/boringssl/src/include/openssl/ssl.h"
 #endif
 
 #if !defined(OS_NACL)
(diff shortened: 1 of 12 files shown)


mojo/public/cpp/platform/secure_socket_utils_posix.cc
```diff
@@ -0,0 +1,111 @@
...
+    CHECK(SSL_CTX_check_private_key(ssl_ctx_.get()));
+  }
+
+  bssl::UniquePtr<SSL_CTX> ssl_ctx_;
+};
+
+class SSLClientContext {
+ public:
+  static SSLClientContext* GetInstance() {
+    return base::Singleton<SSLClientContext,
```
nke94, line 207: Could you explain the reason why it should be a singleton?
Based on this implementation, when two client socket is created and want to make secure connection, it will share same SSL context.
Is it normal way of usage open_SSL library?

mojo/core/channel_posix.cc
```diff
@@ -509,6 +526,10 @@ class ChannelPosix : public Channel,
 
       std::vector<base::ScopedFD> incoming_fds;
       ssize_t read_result =
+#if defined(CASTANETS)
+          secure_connection_ ?
+          SecureSocketRecvmsg(ssl_.get(), buffer, buffer_capacity) :
```
nke94: When secure_connection true.. is vector<base::ScopedFD> incoming_fds handled properly?
If handling incomming_fds_ is not meaningful for secure connection why don't you exclude below [for (auto& fd : incoming_fds)] block in case of secure_connection

mojo/public/cpp/platform/secure_socket_utils_posix.cc
hyowon: We do not need different certificates and different configuration contexts. So whenever we need a secure connection, we can provide an individual SSL instance from a global singleton SSL context.

mojo/core/channel_posix.cc
hyowon: As you say, incoming_fds is not used and empty when secure_connection. I just wanted to make the code diff by defined(CASTANETS) as small and simple as possible.
Doing [for (auto& fd: incoming_fds)] don't burden at all (because it should be empty), so why don't you maintain code simplicity?

hyowon: I've modified what you mentioned.

nke94 (approved):

PR #231, opened 2019-09-19
Apply WaitSync for WebGL.

Sunwoo-Nam:
The WaitSync function guarantees complete data transfer about GUID.


hyowon (approved):
It's surprising that we can handle WaitSync in just one place in GPU related code. LGTM.

Sunwoo-Nam:
> It's surprising that we can handle WaitSync in just one place in GPU related code. LGTM.

I've learned that "Don't trust anyone's source code. Even if it's Google.."
Thank you for your "LGTM"

PR #233, opened 2019-09-19
Fix build error since PR #188.

joonghunpark:
This CL has two changes as below.

1) When build with gn option 'is_component_build = true'
below error occured.

error: duplicate symbol: MojoRetryInvitation

To fix this, use name |MojoRetryInvitationImpl|
instead of |MojoRetryInvitation| at entrypoints.cc.

2) Add additional guards for CASTANETS
which was missed out in PR #188 at thunks.cc.

### DIFF
--- a/mojo/core/entrypoints.cc
+++ b/mojo/core/entrypoints.cc
@@ -354,7 +354,7 @@ MojoResult MojoSendInvitationImpl(
 #endif
 }
 #if defined(CASTANETS)
-MojoResult MojoRetryInvitation(
+MojoResult MojoRetryInvitationImpl(
     const struct MojoPlatformProcessHandle* old_process_handle,
     const struct MojoPlatformProcessHandle* process_handle,
     const struct MojoInvitationTransportEndpoint* transport_endpoint) {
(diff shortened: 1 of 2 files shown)


mojo/core/entrypoints.cc
```diff
@@ -354,7 +354,7 @@ MojoResult MojoSendInvitationImpl(
 #endif
 }
 #if defined(CASTANETS)
-MojoResult MojoRetryInvitation(
+MojoResult MojoRetryInvitationImpl(
```
younghajung, line 357: Please check [line 438](https://github.com/Samsung/Castanets/pull/233/files#diff-a5982dce8dad2b629215942bd6a1a54aR438)

joonghunpark: Thank you for your review:) Done.

mojo/core/broker_castanets.cc
```diff
@@ -415,7 +415,7 @@ void BrokerCastanets::AddSyncFence(const base::UnguessableToken& guid,
                                    uint32_t fence_id) {
   fence_queue_->AddFence(guid, fence_id);
 }
-
+#if defined(CASTANETS)
```
younghajung: I think we don't need the guard in this file, broker_castanets.cc is compiled only when castanets is enabled.

joonghunpark: Hmm, that's why I originally didn't add the flag, but I forgot.
Thank you for your review.

joonghunpark: Done.

younghajung (approved):
LGTM

PR #235, opened 2019-09-20
fixup! Launch renderer process via forking. (#188)

Sunwoo-Nam:
Fix the crash issue if running Castanets with --enable-forking.


joonghunpark (approved):
Ok, the crash occured is the same one we meet if we use OnceCallback instead of RepeatingCallback currently.
LGTM.

PR #237, opened 2019-09-20
Inform the browser of the shared memory created by the renderer.

hyowon:
In case of the command buffer, the renderer creates a shared memory and then share it with the Browser. And both the browser and renderer try to write data onto the shared memory and sync it.
At that point, the browser doesn't know which renderer(node) to send sync data to. Thus, this CL adds a new broker message for informing the browser of the shared memory created by the renderer.


hyowon:
This PR is necessary to run WebGL properly with #231

Sunwoo-Nam (approved):
LGTM

PR #240, opened 2019-09-23
Add --tcp-launch-timeout switch.

joonghunpark:
This switch specifies tcp launch timeout value to fallback
to launch renderer process via forking.

The behavior of this switch is as below.
1) out/Release/chrome --tcp-launch-timeout
Use default timeout value |kTcpLaunchTimeoutDefault|
in child_process_launcher_helper.cc

2) out/Release/chrome --tcp-launch-timeout=[timeout value]
Use the specified timeout value to fall through to launch
renderer process via unix domain socket.

### DIFF
--- a/content/browser/child_process_launcher_helper.cc
+++ b/content/browser/child_process_launcher_helper.cc
@@ -24,9 +24,10 @@
 
 #if defined(CASTANETS)
 #include "base/base_switches.h"
+#include "base/strings/string_number_conversions.h"
 #include "content/browser/renderer_host/input/timeout_monitor.h"
 
-int kTcpConnectionTimeout = 5;
+int kTcpLaunchTimeoutDefault = 10;
 #endif
 
 namespace content {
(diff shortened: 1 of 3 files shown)


content/browser/child_process_launcher_helper.cc
```diff
@@ -96,8 +97,22 @@ ChildProcessLauncherHelper::ChildProcessLauncherHelper(
   relaunch_renderer_process_monitor_timeout_.reset(new TimeoutMonitor(
       base::Bind(&ChildProcessLauncherHelper::OnCastanetsRendererTimeout,
                  base::Unretained(this))));
-  relaunch_renderer_process_monitor_timeout_->Start(
-      base::TimeDelta::FromSeconds(kTcpConnectionTimeout));
+  if (base::CommandLine::ForCurrentProcess()->HasSwitch(
+          switches::kTcpLaunchTimeout)) {
+    int tcp_launch_timeout;
```
Sunwoo-Nam, line 102: How about this?
int tcp_launch_timeout = kTcpLaunchTimeoutDefault;
and unify two different parameters of |timeout_->Start|.
'else' may not be necessary.

joonghunpark: When base::StringToInt gets empty string with only --tcp-launch-timeout switch not being with its timeout value,
then it returns false setting tcp_launch_timeout to '0'.

So in that case if we declare tcp_launch_timeout; as below,

int tcp_launch_timeout = kTcpLaunchTimeoutDefault;

then it will be overriden by 0, I'm afraid.

But if any better solution, please suggest it:)

hyowon (commented):
If the switch doesn't exist, I hope fallback doesn't work.

joonghunpark (commented):
If this switch is not added, fallback will not work as expected.
The commit message describes just 2 cases.
1) --tcp-launch-timeout without explicit timeout value.
2) --tcp-launch-timeout with explicit timeout value.

if (base::CommandLine::ForCurrentProcess()->HasSwitch(
          switches::kTcpLaunchTimeout))
check exists for that purpose.

hyowon (approved):
LGTM

content/browser/child_process_launcher_helper.cc
```diff
@@ -96,8 +97,22 @@ ChildProcessLauncherHelper::ChildProcessLauncherHelper(
   relaunch_renderer_process_monitor_timeout_.reset(new TimeoutMonitor(
```
is46-kim: It needs TimeoutMonitor instance even without the switch?

joonghunpark: Ok, let me move 
relaunch_renderer_process_monitor_timeout_.reset(new TimeoutMonitor)
into
if (base::CommandLine::ForCurrentProcess()->HasSwitch(
          switches::kTcpLaunchTimeout)) {}
condition.

PR #243, opened 2019-09-25
Enable GPU compositing and fix the WebGL issue with --enable-forking.

Sunwoo-Nam:
Enable GPU compositing as a default option for Castanets.
WebGL doesn't run with enable-forking mode because the result is initiated to 0
at the wrong time. So, synchronize the result memory from renderer process.


hyowon (approved):
Cool!

PR #245, opened 2019-09-26
fixup! Inform the browser of the shared memory created by the renderer.

suyambulingamrm:
Use broker channel to send the BUFFER_CREATED (non-sync msg) instead of
synch_channel_. This avoids receiving malformed message on browser side on
different test scenarios.


hyowon (approved):
LGTM

PR #246, opened 2019-09-26
[Workaround] Fix the crash issue caused by OOM.

Sunwoo-Nam:
A new staging buffer is created for every frame because QueryResult is not
handled on Castanets. It causes out-of-memory, so fall-back to glFinish.


Sunwoo-Nam:
This is for a Test.

hyowon (approved):

PR #247, opened 2019-09-27
Fix black tile problem, handle key events specific to tizen platform.

suyambulingamrm:


Sunwoo-Nam (approved):
It looks great. After some tests for verification, we can merge this.

Chikdolman:
Bugfix Verified. LGTM

PR #249, opened 2019-10-02
Fix the deadlock during WebGL loading.

Sunwoo-Nam:
This issue is caused by deadlock with mutex locks.
On GPU thread, after |broker_hosts_lock_| is used, |channel_lock_| is used
to sync memory.
On the other hand, the order is reversed on IO thread to load resources.
This bad situation occurs especially during loading WebGL pages.
Therefore, adjust the scope of locks.

### DIFF
--- a/mojo/core/node_controller.cc
+++ b/mojo/core/node_controller.cc
@@ -349,14 +349,19 @@ bool NodeController::SyncSharedBuffer(
   if (broker_)
     return broker_->SyncSharedBuffer(guid, offset, sync_size);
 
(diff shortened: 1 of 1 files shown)


hyowon (changes requested):
This can be resolved without locking broker_hosts_lock_ in NodeController::Sync*(). I'll create a PR. Thank you for your investigation.

Sunwoo-Nam:
https://github.com/Samsung/Castanets/pull/253

PR #250, opened 2019-10-07
Toggle android specific setting for better renderer performance.

suyambulingamrm:
1. kLaunchingProcessIsBackgrounded is set by default which enables dom timer
throttling, which made SetInterval firing too slow on tizen-android.
2. Enable Periodic flush for android as well.


Sunwoo-Nam:
LGTM

Sunwoo-Nam (approved):

suyambulingamrm:
There is no regressions observed after this change.

PR #252, opened 2019-10-07
Initialize third_party/skia for castanets_69.

Sunwoo-Nam:
third_party/skia integrates with castanets_69.


joonghunpark (approved):
nit: How about import or integrate instead of initialize?

suyambulingamrm:
LGTM.
We need skia module inside castanets to make skia related changes corresponding to respective platforms.

PR #254, opened 2019-10-07
Drop sync IPC messages in chrome/renderer.

suyambulingamrm:
IPC msgs like ChromeViewHostMsg_ContentBlocked,
ChromeViewHostMsg_AllowIndexedDB, ChromeViewHostMsg_AllowDOMStorage
are not handled in platforms like tizen. This leads to loading issues
on major websites, so temporarily dropping these messages.

### DIFF
--- a/chrome/renderer/content_settings_observer.cc
+++ b/chrome/renderer/content_settings_observer.cc
@@ -177,12 +177,16 @@ void ContentSettingsObserver::DidBlockContentType(
 void ContentSettingsObserver::DidBlockContentType(
     ContentSettingsType settings_type,
     const base::string16& details) {
+#if defined(CASTANETS)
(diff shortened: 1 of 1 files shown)


Chikdolman (approved):
In order to drop the msg, IPC Msgs should be investigated further in the future for its actual needs.
However, the change is needed for now for loading major webpages (such as naver, daum, etc).

nke94 (commented):
If possible, it will be better add message handler in platform side.
even though we cannot apply proper feature implement. why don't you add to send expected response message.

suyambulingamrm:
> If possible, it will be better add message handler in platform side.
> even though we cannot apply proper feature implement. why don't you add to send expected response message.

Yes sure. Now we are handling these messages.

PR #255, opened 2019-10-08
Enable media element for castanets

sm-venugopal:
1. Initial patch supporting media element playback
2. Added video hole support for video playback
3. Added video hole preference to WebView
4. Disabled android specific features

### DIFF
--- a/content/renderer/media/castanets/castanets_webmediaplayer_impl.cc
+++ b/content/renderer/media/castanets/castanets_webmediaplayer_impl.cc
@@ -1,0 +1,737 @@
+// Copyright 2019 Samsung Electronics Inc. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "content/renderer/media/castanets/castanets_webmediaplayer_impl.h"
+
+#include "cc/layers/video_layer.h"
+#include "media/base/bind_to_current_loop.h"
+#include "media/base/media_content_type.h"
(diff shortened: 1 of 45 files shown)


joonghunpark:
Which device configuration scenarios does this CL aiming at?
Tizen(browser process) and Tizen, Android(renderer process)?

joonghunpark (commented):
LGTM. But please wait for review from someone others.

sm-venugopal:
> Which device configuration scenarios does this CL aiming at?
> Tizen(browser process) and Tizen, Android(renderer process)?

Thats right.

content/renderer/media/castanets/castanets_webmediaplayer_impl.cc
```diff
@@ -0,0 +1,744 @@
...
+
+size_t WebMediaPlayerCastanets::AudioDecodedByteCount() const {
+  return 0;
+}
+
+size_t WebMediaPlayerCastanets::VideoDecodedByteCount() const {
+  return 0;
+};
+
+bool WebMediaPlayerCastanets::CopyVideoTextureToPlatformTexture(
```
joonghunpark, line 464: This method only returns false.
If this method will have more implementation,
please leave a TODO(). 
Otherwise either 'early return false' or 'NOTIMPLEMENTED();' would be better, I think.

Chikdolman:
Requested to rebuild the Android, as the android build fail occurred due to the network problem from the build server.

sm-venugopal:
Updated with Video hole support for inline playback.

PR #257, opened 2019-10-08
Enable accelerated path for canvas2d when MSAA is used with BGRA format.

suyambulingamrm:
When MSAA is used with BGRA color type, SkSurface creation fails which
make canvas2d fallback to sw path. So this change forcefully allows
SkSurface to be created with BGRA colortype with msaa sample count set.


Chikdolman (approved):
LGTM

PR #265, opened 2019-11-11
Delete temporary local AHardwareBuffer definition.

suyambulingamrm:
This upstream change fixes jumbo build break for android.
Reference: https://chromium-review.googlesource.com/c/chromium/src/+/1141735


younghajung (approved):
LGTM

PR #263, opened 2019-11-07
Build application as docker image

rzr:
I am sharing a code that rely on other PR
that let you build a docker image from sources, as well as a debian package to deploy application more easily:

https://hub.docker.com/r/rzrfreefr/castanets/tags

https://dl.bintray.com/rzr/deb/pool/c/castanets-snapshot/

First please first merge this pending PR
https://github.com/Samsung/Castanets/pull/73

Then I can submit each patches individually.

Regards

PR #269, opened 2019-11-26
docs: Add tips about using docker

rzr:
Change-Id: I957d05025346ecc71e315e687b41d410268add95
Relate-to: https://github.com/Samsung/Castanets/pull/263
Signed-off-by: Philippe Coval <p.coval@samsung.com>


Chikdolman:
I think we need documentation from `docker pull`. However, I'll try to figure out where we can post the docker file.

rzr:
Sure, in next patch I can document docker pull from my dockerhub repo:
https://hub.docker.com/r/rzrfreefr/castanets/tags

PR #271, opened 2019-12-02
Apply zlib,webp based compression for sync data.

suyambulingamrm:
1. Default compression type is zlib.
2. Apply webp for 256*256 tiles.
3. If the sync size is less than 200 bytes, do not apply
   compression.

### DIFF
--- a/cc/raster/one_copy_raster_buffer_provider.cc
+++ b/cc/raster/one_copy_raster_buffer_provider.cc
@@ -362,9 +362,9 @@ void OneCopyRasterBufferProvider::PlaybackToStagingBuffer(
 #if defined(CASTANETS)
     if (std::string("renderer") ==
         base::CommandLine::ForCurrentProcess()->GetSwitchValueASCII("type")) {
-      mojo::SyncSharedMemory(
-          buffer->GetHandle().handle.GetGUID(), 0,
-          staging_buffer->size.width() * staging_buffer->size.height() * 4);
+      mojo::SyncSharedMemory2d(
+          buffer->GetHandle().handle.GetGUID(), staging_buffer->size.width(),
+          staging_buffer->size.height(), buffer->stride(0)/staging_buffer->size.width());
     }
 #endif
     buffer->Unmap();
--- a/mojo/core/broker_castanets.cc
+++ b/mojo/core/broker_castanets.cc
@@ -21,6 +21,12 @@
 #include "mojo/public/cpp/platform/socket_utils_posix.h"
 #include "mojo/public/cpp/platform/tcp_platform_handle_utils.h"
 
+#include "third_party/libwebp/src/webp/decode.h"
+#include "third_party/libwebp/src/webp/encode.h"
+#include "third_party/zlib/google/compression_utils.h"
+
+#define COMPRESSION_THRESHOLD_BYTES 200
+
 namespace mojo {
 namespace core {
 
@@ -267,13 +273,14 @@ void BrokerCastanets::SendSyncEvent(
   CHECK(tcp_connection_);
   SyncSharedBufferImpl(mapping_info->guid(),
                        static_cast<uint8_t*>(mapping_info->GetMemory()), offset,
-                       sync_size, mapping_info->mapped_size(), write_lock);
+                       sync_size, mapping_info->mapped_size(), BrokerCompressionMode::ZLIB, write_lock);
 }
 
 bool BrokerCastanets::SyncSharedBuffer(
     const base::UnguessableToken& guid,
     size_t offset,
-    size_t sync_size) {
+    size_t sync_size,
+    BrokerCompressionMode compression_mode) {
   if (!tcp_connection_)
     return true;
 
@@ -283,7 +290,7 @@ bool BrokerCastanets::SyncSharedBuffer(
       return MOJO_RESULT_NOT_FOUND;
 
   SyncSharedBufferImpl(guid, static_cast<uint8_t*>(mapping->GetMemory()),
-                       offset, sync_size, mapping->mapped_size());
+                       offset, sync_size, mapping->mapped_size(), compression_mode);
   return true;
 }
 
@@ -295,15 +302,17 @@ bool BrokerCastanets::SyncSharedBuffer(
     return true;
 
   SyncSharedBufferImpl(mapping.guid(), static_cast<uint8_t*>(mapping.memory()),
-                       offset, sync_size, mapping.mapped_size());
+                       offset, sync_size, mapping.mapped_size(), BrokerCompressionMode::ZLIB);
   return true;
 }
 
 bool BrokerCastanets::SyncSharedBuffer2d(const base::UnguessableToken& guid,
-                                         size_t offset,
-                                         size_t sync_size,
                                          size_t width,
-                                         size_t stride) {
+                                         size_t height,
+                                         size_t bytes_per_pixel,
+                                         size_t offset,
+                                         size_t stride,
+                                         BrokerCompressionMode compression_mode) {
   if (!tcp_connection_)
     return true;
 
@@ -313,8 +322,8 @@ bool BrokerCastanets::SyncSharedBuffer2d(const base::UnguessableToken& guid,
     return MOJO_RESULT_NOT_FOUND;
 
   SyncSharedBufferImpl2d(guid, static_cast<uint8_t*>(mapping->GetMemory()),
-                         offset, sync_size, mapping->mapped_size(), width,
-                         stride);
+                         mapping->mapped_size(), width, height,
+                         bytes_per_pixel, offset, stride, compression_mode);
   return true;
 }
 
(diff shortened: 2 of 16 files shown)


suyambulingamrm:
Currently only 256*256 tiles are considered for webp compression.
webp encode API needs tile width, height and stride data and we need to pass all data via SyncSharedMemory API in order to compress all sized tiles.
Let me know your inputs.

Currently sync size threshold for compression is taken as 200. Occasionally compressed data size is more than raw data size if sync size is too small (in ~100's)

cc/raster/one_copy_raster_buffer_provider.cc
```diff
@@ -362,9 +362,13 @@ void OneCopyRasterBufferProvider::PlaybackToStagingBuffer(
...
         base::CommandLine::ForCurrentProcess()->GetSwitchValueASCII("type")) {
+      mojo::CompressionMode compression_mode = mojo::CompressionMode::ZLIB;
+      if (staging_buffer->size.width() == 256 &&
+          staging_buffer->size.height() == 256)
+        compression_mode = mojo::CompressionMode::WEBP;
       mojo::SyncSharedMemory(
           buffer->GetHandle().handle.GetGUID(), 0,
-          staging_buffer->size.width() * staging_buffer->size.height() * 4);
+          staging_buffer->size.width() * staging_buffer->size.height() * 4, compression_mode);
     }
```
Chikdolman, line 368: Instead of passing compression_mode(uint32_t), how about passing a class or a structure that includes tile size and types? so that it can cover all tile size

suyambulingamrm: On the latest patch set, we pass the required data as a struct to use webp for all tiles.

mojo/core/broker_castanets.cc
```diff
@@ -323,10 +331,49 @@ void BrokerCastanets::SyncSharedBufferImpl(const base::UnguessableToken& guid,
...
   BufferSyncData* buffer_sync = nullptr;
   void* extra_data = nullptr;
+  uint8_t* start_ptr = static_cast <uint8_t*>(memory + offset);
+  uint8_t* compressed_data = nullptr;
+  int original_sync_size = sync_size;
+  std::string compressed;
+  BrokerCompressionMode compression_mode = compression_data.mode();
+  if (sync_size <= COMPRESSION_THRESHOLD_BYTES)
+    compression_mode = NONE;
+
```
Chikdolman, line 346: Are there cases where compressed size is larger than the original size even for the WebP?

suyambulingamrm: I didn’t observe such case in webp. Only if the original size is too small this case is observed in zlib.  As we use webp only for tiles, it may have sync size of at least 16384 bytes (for 64 size tile).
The quality factor is considered as 100. Should that be ok ?

cc/raster/one_copy_raster_buffer_provider.cc
is46-kim: How about using SyncSharedMemory2d() instead of SyncSharedMemory() for bitmap.
I think we can also support full bitmap(offset is zero and stride is same as width) in SyncSharedMemory2d() .

mojo/core/broker_castanets.cc
Chikdolman: 1. So you mean minimal size is larger than 200 Byte for webP. OK.
2. For WebP, the current code is WebPEncodeRGBA with 100. Why don't we go for WebPEncodeLosslessRGBA? quality factor 100 is still observed poor comparing to lossless.

cc/raster/one_copy_raster_buffer_provider.cc
suyambulingamrm: Ok we can use SyncSharedMemory2d. I have some clarification, 

1. I tried to test SyncSharedMemory2d for partial tiles, but it looks the module is not working properly, it looks stride/width is not proper. Only after I modify as below I could see proper unbroken tiles, else data corruption happens. Can you please clarify whether my understanding is wrong.
https://github.com/suyambulingamrm/Castanets/commit/c881318e3914c05be96ab741f54689502f46501a

2. full bitmap(offset is zero and stride is same as width) => stride (rowbytes) should be width * bytesperpixel ?

is46-kim: https://github.com/Samsung/Castanets/pull/205/commits/c9f55f1b755f7eb4a4464867d5de464ad984f2f1
Yes, Sunwoo intended width, height, offset as bytes, not pixel.
Currently, this function seems to be not used anymore. How about changing function parameter for webp compression?

suyambulingamrm: Oh ok.. Changing the Sync2d API as mentioned below is ok ? we need height info of the rect as well for webp encode api.
(From) SyncSharedMemory2d(const base::UnguessableToken& guid,
                   size_t offset,
                   size_t sync_size,
                   size_t width,
                   size_t stride,
                   BrokerCompressionData compression_data=BrokerCompressionData());
(To) SyncSharedMemory2d(const base::UnguessableToken& guid,
                   size_t offset,
  		   size_t width,
  		   size_t height,
                   size_t bytesperpixel,
                   size_t stride,
		   CompressionMode mode);

is46-kim: LGTM. But I prefer the following order : size_t width, size_t height, size_t bpp, size_t offset, size_t stride, CompressionMode mode
And I think pixel is better than byte for width, height, offset, and stride.

mojo/core/broker_castanets.cc
suyambulingamrm: Yes right, I have updated it to use lossless api.

cc/raster/one_copy_raster_buffer_provider.cc
suyambulingamrm: Updated on latest patchset to use 2d api for tiles as suggested.

mojo/core/broker_castanets.cc
```diff
@@ -463,19 +572,69 @@ void BrokerCastanets::OnBufferSync2d(uint64_t guid_high,
...
-           static_cast<const uint8_t*>(data) + put_bytes, width);
-    put_offset += stride;
-    put_bytes += width;
+  if (mapping) {
+    CHECK_GE(mapping->mapped_size(), offset + sync_bytes);
+    size_t put_bytes = 0;
+    size_t put_offset = offset;
+    if (offset == 0 && stride == 0) {
+      // Full tile
+      memcpy(static_cast<uint8_t*>(mapping->GetMemory()) + offset,
```
is46-kim: It seems to compress data directly for full tile without copy

mojo/core/broker_castanets.cc
```diff
@@ -353,13 +392,62 @@ void BrokerCastanets::SyncSharedBufferImpl(const base::UnguessableToken& guid,
...
+  size_t sync_size = width * height * bytes_per_pixel;
   CHECK_GE(mapped_size, offset + sync_size);
+  size_t put_bytes = 0;
+  size_t put_offset = offset;
+  uint32_t buffer_stride = width * bytes_per_pixel;
+
+  std::vector<uint8_t> bytes(sync_size);
+  if (offset == 0 && stride == 0) {
+      // Full tile
+      memcpy(static_cast<uint8_t*>(&bytes[0]) + offset, memory,
```
is46-kim: It seems we can compress the memory directly without copy.

mojo/core/broker_castanets.cc
```diff
@@ -463,19 +572,69 @@ void BrokerCastanets::OnBufferSync2d(uint64_t guid_high,
...
+
+  switch (compression_mode) {
+    case BrokerCompressionMode::ZLIB: {
+      LOG(ERROR) << "Unsupported compression mode";
+      break;
+    }
+    case BrokerCompressionMode::WEBP: {
+      int width = 0;
+      int height = 0;
+      uint8_t* result = WebPDecodeRGBA(
```
is46-kim: It seems to need free..
// The returned pointer should be deleted calling WebPFree().

is46-kim: It seems that we can decode directly to shared memory without intermediate buffer.
WEBP_EXTERN uint8_t* WebPDecodeRGBAInto(
    const uint8_t* data, size_t data_size,
    uint8_t* output_buffer, size_t output_buffer_size, int output_stride);

mojo/core/broker_castanets.cc
suyambulingamrm: Updated the patchset without copy.

mojo/core/broker_castanets.cc
suyambulingamrm: Updated the patchset to decode directly to shared memory. Kindly take a look.
WebPFree() is used to delete the encoded result ptr.

mojo/core/broker_castanets.cc
```diff
@@ -353,13 +392,65 @@ void BrokerCastanets::SyncSharedBufferImpl(const base::UnguessableToken& guid,
...
+    case BrokerCompressionMode::ZLIB: {
+      LOG(ERROR) << "Unsupported compression mode";
+      break;
+    }
+    case BrokerCompressionMode::WEBP: {
+      size_t webp_stride = stride ? stride : buffer_stride;
+      start_ptr = memory + offset;
+      size_t size;
+      size = WebPEncodeLosslessRGBA(start_ptr, width,
+          height, webp_stride, &compressed_data);
```
is46-kim, line 425: Needs WebPFree() for |compressed_data| to reclaim memory?

suyambulingamrm: I have added it in Line 474.  It seems to be ok ?
 if (compression_mode == BrokerCompressionMode::WEBP && compressed_data)
    WebPFree (compressed_data);

mojo/core/broker_castanets.cc
```diff
@@ -463,19 +578,74 @@ void BrokerCastanets::OnBufferSync2d(uint64_t guid_high,
...
+    base::SharedMemoryCreateOptions options;
+    options.size = buffer_bytes;
+    base::subtle::PlatformSharedMemoryRegion handle =
+        base::CreateAnonymousSharedMemoryIfNeeded(guid, options);
+    CHECK(handle.IsValid());
+
+    void* memory = mmap(NULL, original_size + offset, PROT_READ | PROT_WRITE,
+                        MAP_SHARED, handle.GetPlatformHandle().fd, 0);
+    ptr = static_cast<uint8_t*>(memory);
+  }
```
is46-kim, line 596: How about initializing |ptr| as |mapping->GetMemory()| even if |mapping| in null?
It seems to remove codes for |mapping| in line 607, 629.

mojo/core/broker_castanets.cc
is46-kim: Looks good:)

mojo/core/broker_castanets.cc
suyambulingamrm: Yes, we can do that.. Updated in latest patchset.

is46-kim (approved):
LGTM

PR #273, opened 2019-12-23
Modify TCP connection for the audio feature.

Sunwoo-Nam:
Renderer process can be TCP server since the below patch.
"Support both server and client TCP sockets on all processes"

So the audio feature also supports both server and client TCP sockets
on all processes.

Signed-off-by: Sunwoo Nam <sunny.nam@samsung.com>

### DIFF
--- a/content/renderer/media/audio/mojo_audio_output_ipc.cc
+++ b/content/renderer/media/audio/mojo_audio_output_ipc.cc
@@ -12,6 +12,8 @@
 #include "mojo/public/cpp/system/platform_handle.h"
 
 #if defined(CASTANETS)
+#include "base/base_switches.h"
+#include "base/command_line.h"
 #include "mojo/public/cpp/platform/tcp_platform_handle_utils.h"
 #endif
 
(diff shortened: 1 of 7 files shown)


is46-kim (approved):

content/renderer/media/audio/mojo_audio_output_ipc.cc
```diff
@@ -260,16 +278,30 @@ void MojoAudioOutputIPC::Created(
...
     base::UnsafeSharedMemoryRegion shared_memory_region,
-    int32_t port) {
-  // Create a TCP client socket.
-  mojo::PlatformHandle tcp_client_handle = mojo::CreateTCPClientHandle(port);
-  if (!tcp_client_handle.is_valid()) {
-    LOG(ERROR) << __func__ << " tcp_client_handle is not valid.";
-    return;
+    mojo::PlatformHandle server_handle,
+    uint16_t port) {
+  bool is_tcp_server = !port ? true : false;
```
joonghunpark: It seems that "!" mark isn't needed here.

I guess you mean,
bool is_tcp_server = port ? true : false;

media/mojo/services/mojo_audio_output_stream.cc
```diff
@@ -67,16 +69,39 @@ void MojoAudioOutputStream::SetVolume(double volume) {
...
-  uint16_t port;
-  // Create a server TCP socket and get the random port number.
-  mojo::PlatformHandle server_handle = mojo::CreateTCPServerHandle(0, &port);
-  // Ack with the port number.
-  std::move(callback).Run(port);
-
-  base::ScopedFD accept_handle;
-  mojo::TCPServerAcceptConnection(server_handle.GetFD().get(), &accept_handle);
-  delegate_->OnTCPConnected(accept_handle.release());
+  bool is_tcp_server = !port ? true : false;
```
joonghunpark: Ditto.

content/renderer/media/audio/mojo_audio_output_ipc.cc
Sunwoo-Nam: If  "!" is removed, this will be changed to "bool is_tcp_server = port ? false : true;"

joonghunpark: For example, line 250, stream_->RequestTCPConnect(0, ...) is being called, 
and this call is in 
if (base::CommandLine::ForCurrentProcess()->HasSwitch(switches::kServerAddress)
block,
so I think, this call is for tcp client.
But with port == 0, 
bool is_tcp_server = !port ? true : false;
will set is_tcp_server = true.
So I'm confused a little.

joonghunpark: Hmm, this call was through mojo. So I've mis-read this thing. Sorry for your inconvenience.

joonghunpark (approved):

PR #274, opened 2019-12-24
fixup! [DWE] Fixed SVACE and Coverity issues

djmixkim:
All memeber of socket are will created or deleted only in ctor/dtor.
This patch removes redundant codes.

Signed-off-by: DongJun Kim <djmix.kim@samsung.com>


younghajung (approved):
LGTM

PR #275, opened 2020-01-03
[meerkat] Implement authentication process via Google account

younghajung:
This patch implements the authentication process via Google account for
meerkat server and client. In order to exchange Google ID token, makes
that TCP socket can be used with SSL.
Plus, synchronizes code with Chromium-efl of Tizen.


third_party/meerkat/Component/mmOSAL/timeAPI.cpp
```diff
@@ -40,14 +39,14 @@ OSAL_Time_Return __OSAL_TIME_GetTimeMS(UINT64* ptimeval) {
 #elif defined(LINUX)
   struct timeval tv;
   gettimeofday(&tv, NULL);
-  *ptimeval = ((UINT64)(tv.tv_sec) * 1000) + ((UINT64)(tv.tv_usec) / 1000);
+  *ptimeval = ((UINT64)tv.tv_sec * 1000) + ((UINT64)tv.tv_usec / 1000);
   return OSAL_Time_Success;
 #endif
```
Chikdolman, line 44: Do we use this function also in Android? If so, don't we need a define for Anroid?

younghajung: Yes, we use this in Android as well, the build config of Android defines both "LINUX" and "ANDROID".

Chikdolman (approved):
LGTM

PR #276, opened 2020-01-15
Add switches by default

is46-kim:
- Wait 1 min for TCP client connection
- Add |--disable-frame-rate-limit| switch for Tizen


joonghunpark (approved):
LGTM.

PR #297, opened 2020-05-13
fixup! Support both server and client TCP sockets on all processes.

suyambulingamrm:
Need to mark transport endpoint's secure_connection to true when ssl
based connection is used for correct behavior of --secure-connection
runtime flag.


hyowon (approved):
Okay.

Sunwoo-Nam (approved):
Good. Thank you

PR #300, opened 2020-07-22
fixup! Allow the  node channel to connect using ssl

uzair-jaleel:
This patch corrects few lines of code which were
misplaced during initial feature bringup

Signed-off-by: Uzair Jaleel <uzair.jaleel@samsung.com>


suyambulingamrm (commented):
+1.
The fix is already landed in chromium-efl

suyambulingamrm (approved):
Fix for misplaced lines during bringup.

PR #301, opened 2020-07-22
fixup! Make IPC params similar for all target profiles.

suyambulingamrm:
This change,
1. Reorder WebPreferences struct order same as efl port.
2. Adds a mojo IPC param in CompositorFrameMetadata to match with
   efl port.


suyambulingamrm:
tizen-android has deserialization issue on latest chromium-efl. This fixes the same..

Sunwoo-Nam (approved):
Looks good to me.

PR #308, opened 2020-08-28
[Service Offloading] Enable UserMediaScreenCapturing for Android.

Sunwoo-Nam:
The android device shares own screen to the client device.
UserMediaScreenCapturing option, therefore, should be set to default true.


Chikdolman (approved):
The changes LGTM

PR #309, opened 2020-08-31
Fix build error in android with castanets disabled.

suyambulingamrm:


Chikdolman (approved):
Fixes build error without castanets flag.
LGTM

PR #312, opened 2020-09-02
Enable scrambled message IDs for castanets.

suyambulingamrm:
Mojo message scrambling was initially disabled for castanets in order
to match the IDs between desktop and android, as in android the feature
is disabled. Nacl depends on message scrambling and scrambling enhances
security. So we enable it for castanets too.

Reference:
https://codereview.chromium.org/2794743002/


is46-kim (approved):
LGTM

PR #313, opened 2020-09-02
fixup! Fix build errors in android with castanets disabled.

suyambulingamrm:
This change fixes the build error on desktop with castanets disabled.

build is verified with below config,
linux desktop: with, without castanets build option.
android: with, without castanets build option.


Chikdolman (approved):
LGTM

yg48 (commented):
RGTM

PR #314, opened 2020-09-03
Adds variable for Castanets to AndroidManifest.

Sunwoo-Nam:
Java codes for android need to know if Castanets is enabled.
enable_castanets is added to the manifest for this case.

It can be determined whether the activity is to be hidden for Castanets.

### DIFF
--- a/chrome/android/java/src/org/chromium/chrome/browser/init/AsyncInitializationActivity.java
+++ b/chrome/android/java/src/org/chromium/chrome/browser/init/AsyncInitializationActivity.java
@@ -7,6 +7,9 @@ package org.chromium.chrome.browser.init;
 import android.annotation.SuppressLint;
 import android.content.Context;
 import android.content.Intent;
+import android.content.pm.ApplicationInfo;
+import android.content.pm.PackageManager;
+import android.content.pm.PackageManager.NameNotFoundException;
 import android.content.res.Configuration;
 import android.os.Build;
 import android.os.Bundle;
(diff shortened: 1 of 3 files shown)


chrome/android/java/src/org/chromium/chrome/browser/init/AsyncInitializationActivity.java
```diff
@@ -269,8 +273,19 @@ public void onStartupFailure() {
     @Override
     @SuppressLint("MissingSuperCall")  // Called in onCreateInternal.
     protected final void onCreate(Bundle savedInstanceState) {
-        boolean runRenderer = true; // Toggle this when running browser process
-        if (runRenderer) {
+        boolean isCastanets = false; // Toggle this when running browser process
+
+        Context context = ContextUtils.getApplicationContext();
+        ApplicationInfo info;
```
yg48: How about moving declaration of  |ApplicationInfo| to "try" statement?
|info| variable isn't used anywhere except for "try" statement.

Chikdolman (approved):
Runtime flag added for castanets. LGTM

chrome/android/java/src/org/chromium/chrome/browser/init/AsyncInitializationActivity.java
Sunwoo-Nam: Done. thank you

yg48 (commented):
LGTM :)

yg48 (approved):

PR #315, opened 2020-09-03
[Service Offloading] Check REQUEST_AUDIO permission in Service Offloa…

Sunwoo-Nam:
…ding only.

Decide whether to work with permission checks using the variable of Manifest.


Chikdolman (approved):
Modification LGTM

PR #319, opened 2020-09-06
Fix dispatch error for FrameHostMsg_FocusedNodeChanged IPC.

suyambulingamrm:
IPC dispatch error is observed for FrameHostMsg_FocusedNodeChanged
when we focus button/textbox. This change makes the params similar to
efl port and also adds a error log when such dispatch error occurs
for debugging.


Chikdolman (approved):
Fixes bad message error when clicking text box or button.

PR #321, opened 2020-09-09
[Service Offloading] Refactor java codes using build macros.

Sunwoo-Nam:
1. Add OffloadingUtils.java to check build macros.
2. Retrieve removed java codes for the compatibility issue. (refectoring)
3. Code formatting.

### DIFF
--- a/media/capture/content/android/java/src/org/chromium/media/ImageWrapper.java
+++ b/media/capture/content/android/java/src/org/chromium/media/ImageWrapper.java
@@ -3,7 +3,6 @@ package org.chromium.media;
 import android.annotation.SuppressLint;
 import android.graphics.ImageFormat;
 import android.graphics.PixelFormat;
-import android.graphics.Rect;
 import android.media.Image;
 import android.util.Log;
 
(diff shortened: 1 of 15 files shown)


media/capture/content/android/java/src/org/chromium/media/ImageWrapper.java
```diff
@@ -12,47 +11,49 @@
...
+                                    + image.getPlanes().length);
                     throw new IllegalStateException();
-            }
+                }
+                ScreenCaptureService.sVideoBuffer[0].rewind();
+                ScreenCaptureService.sVideoBuffer[0].put(image.getPlanes()[0].getBuffer());
+                rowStride[0] = image.getPlanes()[0].getRowStride();
+                break;
+            case ImageFormat.YUV_420_888:
+                if (image.getPlanes().length != 3) {
```
yg48: How about using const variable like |number_of_planes_for_YUV420| instead of 3?

media/capture/content/android/java/src/org/chromium/media/ScreenOffloading.java
```diff
@@ -0,0 +1,327 @@
...
+        String TAG = "[NSW] MyBroadcastReceiver";
+        private final WeakReference<ScreenOffloading> mWeakParent;
+        public MyBroadcastReceiver(final ScreenOffloading parent) {
+            mWeakParent = new WeakReference<ScreenOffloading>(parent);
+        }
+
+        @Override
+        public void onReceive(final Context context, final Intent intent) {
+            final String action = intent.getAction();
+            if (ScreenCaptureService.ACTION_QUERY_STATUS_RESULT.equals(action)) {
```
yg48: This condition is needed?

media/capture/content/android/java/src/org/chromium/media/ScreenOffloading.java
```diff
@@ -0,0 +1,327 @@
...
+            byte[] temp = new byte[ScreenCaptureService.sAudioBuffer.remaining()];
+            synchronized (ScreenCaptureService.sAudioSync) {
+                ScreenCaptureService.sAudioBuffer.position(0);
+                ScreenCaptureService.sAudioBuffer.get(temp);
+
+                ScreenCaptureService.sAudioSync.notifyAll();
+            }
+
+            AudioRecordBridge.getAudioRecord().sAudioBuffer.clear();
+            AudioRecordBridge.getAudioRecord().sAudioBuffer.put(temp);
```
yg48: In |put| function, |temp|'s data is copied to |sAudioBuffer|'s buffer?
If |sAudioBuffer| has its own buffer, |temp| needs to be allocated every time?

media/capture/content/android/java/src/org/chromium/media/ImageWrapper.java
Sunwoo-Nam: Actually I referenced previous codes, but I think your comment is better.

media/capture/content/android/java/src/org/chromium/media/ScreenOffloading.java
Sunwoo-Nam: No. Thanks.

media/capture/content/android/java/src/org/chromium/media/ScreenOffloading.java
Sunwoo-Nam: Updated. Thanks.

yg48 (approved):

PR #323, opened 2020-09-10
Show home screen intent only when castanets runtime flag is provided.

suyambulingamrm:
This change,
i. Shows home screen intent only when --enable-castanets is provided.
ii. Corrects README with updated runtime flags.

### DIFF
--- a/chrome/android/java/src/org/chromium/chrome/browser/init/AsyncInitializationActivity.java
+++ b/chrome/android/java/src/org/chromium/chrome/browser/init/AsyncInitializationActivity.java
@@ -31,6 +31,8 @@ import android.view.ViewTreeObserver.OnPreDrawListener;
 import android.view.WindowManager;
 
 import org.chromium.base.ApiCompatibilityUtils;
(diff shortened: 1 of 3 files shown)


chrome/android/java/src/org/chromium/chrome/browser/init/AsyncInitializationActivity.java
```diff
@@ -280,10 +282,12 @@ public void onStartupFailure() {
     protected final void onCreate(Bundle savedInstanceState) {
         // Hide the activity if castanets is enabled and runs as renderer process,
         if (OffloadingUtils.IsCastanets()) {
```
Sunwoo-Nam: Very good.
I think OffloadingUtils.IsCastanets() isn't necessary anymore.
Could you remove this condition?

suyambulingamrm: Done.

Sunwoo-Nam (approved):
LGTM

PR #327, opened 2020-09-17
[Service Offloading] Supports input handling

Chikdolman:
Support Key event and mouse event
For Android, only Galaxy phone is enabled with the patch.
Even for Galaxy, the Knox library should be manually added


Sunwoo-Nam (approved):
LGTM

PR #329, opened 2020-09-21
[Service Offloading] Android Input with Knox Activation Key Support

Chikdolman:
We split AsyncInitializationActivity for two cases.
1. When service_offloading is enabled, we use AsyncInitializationActivityKnox.java
2. When service_offloading is disabled, we use AsyncInitializationActivity.java
AsyncInitializationActivityKnox.java supports Knox activation key

### DIFF
--- a/.gn
+++ b/.gn
@@ -678,4 +678,7 @@ exec_script_whitelist =
       "//remoting/host/installer/win/generate_clsids.gni",
 
       "//tools/grit/grit_rule.gni",
+
+      #Service Offloading: Dynamically copies java file at compiletime
+      "//chrome/android/chrome_java_sources.gni",
     ]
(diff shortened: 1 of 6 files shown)


yg48 (commented):
Only difference is checking key part?
Then how about making sub class for this?
We can apply template method pattern.
https://en.wikipedia.org/wiki/Template_method_pattern

Chikdolman:
> Only difference is checking key part?
> Then how about making sub class for this?
> We can apply template method pattern.
> https://en.wikipedia.org/wiki/Template_method_pattern

This patch enables compile-time build flag. There could be a better way. I'll check once again.

Sunwoo-Nam (approved):
LGTM

Chikdolman:
Knox activation added

PR #332, opened 2020-10-12
Keep Knox sdk APIs to avoid Proguard optimization.

Sunwoo-Nam:
This fixes the crash issue with runtime exception "Stub!".
It's caused by proguard, so add proguard rules for knox sdk.


yg48 (approved):
Good!!

Chikdolman:
verified.

PR #334, opened 2020-10-13
Keep screen on if the castanets activity is shown.

Sunwoo-Nam:
When the screen is turned off, performance may be degraded.
Add the keepScreenOn option and description on the activity.


Sunwoo-Nam:
If anyone knows how to keep the screen on without an activity, please comment here.

Sunwoo-Nam:
I updated the commit for keeping the screen on without the activity.

Chikdolman (approved):
Great! now the device won't sleep

PR #339, opened 2020-10-16
[Service Offloading] Metakey Support

Chikdolman:
Shift, Ctrl, Alt needs to be handled in the code.
This patch supports key events with meta keys


Sunwoo-Nam (approved):
Looks good to me.
It should be checked that the meta keys can be executed standalone.

PR #342, opened 2020-10-20
Fix tile broken issue when |castanets| option is on.

yg48:
Tiles are encoded by |WebPEncodeLosslessRGBA| function when |castanets| option
is on. Although |WebPEncodeLosslessRGBA| is lossless encoding function,
original data could be changed for optimization if tile has transparent area.
To prevent this optimization, set |exact| property as 1.

Signed-off-by: YongGeol Jung <yg48.jung@samsung.com>


Sunwoo-Nam (approved):
Looks good to me.

Chikdolman (approved):

PR #343, opened 2020-10-20
[Service Offloading] Fix the logic of screen capturing to fix crash.

Sunwoo-Nam:
Android requires user consent for screen capturing every time.
This commit improves the user content-related codes
to fix crash and improve usability.


yg48 (approved):
LGTM!

PR #344, opened 2020-10-20
[Service Offloading] Additional API Support to Kill Application

Chikdolman:
window.input_control.stopApplication("pkgName") is supported.
The API kills pkg of the Android application.
When terminating application from remote, stopApplication should be called.


Sunwoo-Nam:
Isn't there a build break in Windows build?

Chikdolman:
> Isn't there a build break in Windows build?

You are right. Fixed code for Windows build.

Sunwoo-Nam (approved):
LGTM.
It would be better to change the name of the module later.
ex) input control -> app control

PR #345, opened 2020-10-21
Fix the black screen after setting the overlay permission.

Sunwoo-Nam:
Create a new setting activity for handling the overlay permission.

### DIFF
--- a/chrome/android/chrome_java_sources.gni
+++ b/chrome/android/chrome_java_sources.gni
@@ -1830,6 +1830,7 @@ chrome_java_sources = [
 
 chrome_java_sources += [
   "java/src/org/chromium/chrome/browser/AlwaysOnTopService.java",
+  "java/src/org/chromium/chrome/browser/CastanetsSettings.java",
   "java/src/org/chromium/chrome/browser/firstrun/CastanetsFragment.java",
   "java/src/org/chromium/chrome/browser/LicenseAdapter.java",
   "java/src/org/chromium/chrome/browser/SampleAdminReceiver.java",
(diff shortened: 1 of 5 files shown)


yg48 (approved):
LGTM. 👍

chrome/android/chrome_java_sources.gni
```diff
@@ -1830,6 +1830,7 @@ chrome_java_sources = [
 
 chrome_java_sources += [
   "java/src/org/chromium/chrome/browser/AlwaysOnTopService.java",
+  "java/src/org/chromium/chrome/browser/CastanetsSettings.java",
   "java/src/org/chromium/chrome/browser/firstrun/CastanetsFragment.java",
```
Chikdolman, line 1834: Should this be added only when Castanets flag is on?

chrome/android/chrome_java_sources.gni
```diff
@@ -1830,6 +1830,7 @@ chrome_java_sources = [
 
 chrome_java_sources += [
   "java/src/org/chromium/chrome/browser/AlwaysOnTopService.java",
+  "java/src/org/chromium/chrome/browser/CastanetsSettings.java",
```
Chikdolman, line 1833: Should this be added only when Castanets flag is on?

Chikdolman (approved):
LGTM we can refactor later

chrome/android/chrome_java_sources.gni
Sunwoo-Nam: You're right but we should use dynamic loading for solving this problem.
I'm going to refactor later... :)

PR #347, opened 2020-10-23
[Service Offloading] Minimize after start screen capture

Chikdolman:
Minimize activity after allow screen capture
If disallow, the activity is terminated only for Service Offloading

### DIFF
--- a/media/capture/content/android/java/src/org/chromium/media/ScreenOffloading.java
+++ b/media/capture/content/android/java/src/org/chromium/media/ScreenOffloading.java
@@ -26,6 +26,7 @@ import android.widget.Toast;
 import org.chromium.base.ApplicationStatus;
 import org.chromium.base.AudioRecordBridge;
 import org.chromium.base.ContextUtils;
+import org.chromium.base.OffloadingUtils;
 import org.chromium.base.Log;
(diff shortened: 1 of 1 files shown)


media/capture/content/android/java/src/org/chromium/media/ScreenOffloading.java
```diff
@@ -193,11 +194,16 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
             mResultCode = resultCode;
             mResultData = data;
             changeCaptureStateAndNotify(CaptureState.ALLOWED);
+            if (OffloadingUtils.IsServiceOffloading()) {
+                ApplicationStatus.getLastTrackedFocusedActivity().moveTaskToBack(true);
```
Sunwoo-Nam, line 198: Just my opinion. How about notify that screen capture is started using Toast?

Chikdolman: @Sunwoo-Nam  We have screen recording msg on top of android once recording starts. Therefore, I think we don't need toast.

Sunwoo-Nam (approved):
LGTM

PR #349, opened 2020-10-26
[Service Offloading] Additional API Support for Start Application

Chikdolman:
window.input_control.startApplication("pkgName") is supported.
The API starts Android pkg.


yg48 (approved):
LGTM!

PR #351, opened 2020-10-28
Launch tasks separately for Offload Settings.

Sunwoo-Nam:
When Offload Settings is launched, SettingsActivity should be shown but
ChromeTabbedActivity is shown unexpectedly.
For solving the issue, use android:documentLaunchMode="intoExisting"
to launch a separate task based on the intent's Component name and data URI.
Without this, the activity will share all the same activities.


is46-kim (approved):
LGTM!

PR #352, opened 2020-10-28
[Service Offloading] Support Wifi Address API from JS

Chikdolman:
We need Wifi as a key for loading application from server.
Therefore we support wifi address function from JS.


kenxiniz:
Could you please set the ip to "127.0.0.1" when network is off.

Chikdolman:
> Could you please set the ip to "127.0.0.1" when network is off.

Done

kenxiniz (approved):
supporting ip address on javascript layer

Sunwoo-Nam (approved):
Looks good to me

PR #355, opened 2020-11-03
[Service Offloading] Fix the frame size issue caused by the orientation.

Sunwoo-Nam:
Fix the workaround codes for the orientation and device change.
This patch make the video fill the screen on any device.

### DIFF
--- a/media/capture/content/android/java/src/org/chromium/media/ScreenCaptureService.java
+++ b/media/capture/content/android/java/src/org/chromium/media/ScreenCaptureService.java
@@ -53,6 +53,13 @@ public class ScreenCaptureService extends Service {
         int STOPPED = 4;
     }
 
+    @IntDef({DeviceOrientation.PORTRAIT, DeviceOrientation.LANDSCAPE})
+    @Retention(RetentionPolicy.SOURCE)
+    private @interface DeviceOrientation {
+        int PORTRAIT = 0;
+        int LANDSCAPE = 1;
+    }
+
     private final Object mCaptureStateLock = new Object();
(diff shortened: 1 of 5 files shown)


media/capture/content/android/java/src/org/chromium/media/ScreenCaptureService.java
```diff
@@ -384,4 +411,61 @@ private void showNotification(final CharSequence text) {
...
+        result.setAction(ACTION_ROTATE_RESULT);
+        result.putExtra(EXTRA_RESULT_ROTATION, rotation);
+        sendBroadcast(result);
+        return true;
+    }
+
+    private void rotateCaptureOrientation(@DeviceOrientation int orientation) {
+        if ((orientation == DeviceOrientation.LANDSCAPE && mWidth < mHeight)
+                || (orientation == DeviceOrientation.PORTRAIT && mHeight < mWidth)) {
+            mWidth += mHeight - (mHeight = mWidth);
```
Chikdolman, line 468: Minor question.   
I just wonder if there would be a case for ``orientation == DeviceOrientation.LANDSCAPE && mWidth > mHeight``?
Wouldn't one of the option (``orientation == DeviceOrientation.LANDSCAPE`` or ``mWidth < mHeight``) enough?
Same for PORTRAIT

Sunwoo-Nam: orientation == DeviceOrientation.LANDSCAPE && mWidth < mHeight
The above condition is meaning that the rotation has occurred because mWidth should be larger then mHeight in the landscape mode.

Chikdolman (approved):

PR #358, opened 2020-11-10
Change the Service Offloading build flag to the runtime flag.

Sunwoo-Nam:
Castanets and Service Offloading should be switched in runtime.


younghajung (approved):
LGTM

PR #359, opened 2020-11-11
[Service Offloading] Add activity for integrated settings

younghajung:
This patch adds activity for integrated settings.
After the patch, neet to run build/update_internal_repo.sh again.


Sunwoo-Nam (approved):
Looks good to me.

Sunwoo-Nam (approved):
Looks good to me.
I think you don' need to add a [Service Offloading] title for the second commit.

PR #360, opened 2020-11-11
Always hide the activity while Castanets is running.

Sunwoo-Nam:
The activity of Castanets causes black screen issue when another activity is
on the castanets and then disappears.
This patch makes the activity of Castanets be always hidden and
removes the unnecessary fragment.


chrome/android/java/src/org/chromium/chrome/browser/firstrun/FirstRunActivity.java
```diff
@@ -520,4 +526,16 @@ public static void setObserverForTest(FirstRunActivityObserver observer) {
...
     }
+
+    private void checkPermission() {
+        if (!Settings.canDrawOverlays(this)) {
+            Intent intent = new Intent(FirstRunActivity.this, CastanetsSettings.class);
+            startActivity(intent);
+            finish();
+        } else {
+            startService(new Intent(this, AlwaysOnTopService.class));
+            moveTaskToBack(true);
```
Chikdolman, line 537: Is there a possibility that moveTaskToBack will cause the page loading performance issue?

Sunwoo-Nam: No issue.
Castanets is floating a transparent view in AlwaysOnTopService.
The code is just above moveTaskToBack..
   startService(new Intent(this, AlwaysOnTopService.class));

Chikdolman (approved):

PR #361, opened 2020-11-11
[Meerkat] Start Meerkat via Offload Settings

younghajung:
This patch adds startService and stopService method to
MeerkatServerService, and removes code that launchs Meerkat in
LaunchIntentDispatcher.
After this, Meerkat will be started via Offload Settings.


is46-kim (approved):
LGTM

PR #362, opened 2020-11-12
[Service Offloading] Fix the getUserMedia issue.

Sunwoo-Nam:
getUserMedia doesn't work because of the audio capture path.
This patch separates the audio input path between system capture and the voice.
Also, the SingleProcess mode causes the issue that the video from the camera
using getUserMedia is not shown. So the SingleProcess option is removed.

### DIFF
--- a/media/base/audio_parameters.cc
+++ b/media/base/audio_parameters.cc
@@ -7,6 +7,10 @@
 #include "base/logging.h"
 #include "media/base/limits.h"
 
+#if defined(SERVICE_OFFLOADING)
+#include "base/distributed_chromium_util.h"
+#endif
+
 namespace media {
 
 base::CheckedNumeric<uint32_t> ComputeAudioInputBufferSizeChecked(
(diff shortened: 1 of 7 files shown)


media/base/audio_parameters.cc
```diff
@@ -93,6 +97,10 @@ void AudioParameters::Reset(Format format,
   frames_per_buffer_ = frames_per_buffer;
   effects_ = NO_EFFECTS;
   mic_positions_.clear();
+#if defined(SERVICE_OFFLOADING)
+  if (base::ServiceOffloading::IsEnabled())
+    capture_system_audio_ = false;
```
Chikdolman, line 102: Since we have set_capture_system_audio function, I think It would be ``set_capture_system_audio(false)``. and make ``capture_system_audio_`` private. If you do not want to make it private, I don't think we need ``set_capture_system_audio``

Sunwoo-Nam: |capture_system_audio_| is already a private variable.
And a code what you commented is for initialization. In this case, other variables also don't use the 'Setter' function.

Chikdolman (approved):

PR #363, opened 2020-11-12
Add paramter for getting "startService" method

djmixkim:
Parameter of startService fucntion is changed in offloadWorker.
This patch fixes parameter for getting corresponding method.


is46-kim (approved):
LGTM

younghajung (approved):

PR #365, opened 2020-11-13
[Service Offloading] Crash with getDisplayMedia constraints except fo…

Sunwoo-Nam:
…r audio.

If getDisplayMedia is used with the constraints which includes only 'video'
except for 'audio', it makes the crash issue because an audio stream device
is always added now.
This patch adds a condition to check whether audio is requested or not.


Chikdolman (approved):
👍

PR #366, opened 2020-11-13
[Meerkat] Implement writing capability

younghajung:
This patch implements writing capability on Meekat.
The application can write the capability as follows.

ContentValues value = new ContentValues();
value.put("capability", "application capability");
try {
  getContentResolver().insert(
      Uri.parse("content://com.samsung.android.meerkat.CapabilityProvider"),
      value);
} catch (Exception e) {
}


djmixkim (approved):
lgtm

PR #367, opened 2020-11-16
[Service Offloading] Move permission check to Offload Settings.

Sunwoo-Nam:
Service Offloading requires three permissions.
Audio recording, Write external storage, Knox.
To improve usability, move permission check to Offload Settings application.


Chikdolman (approved):
LGTM

PR #369, opened 2020-11-18
[Meerkat] Use multicast address only for service offloading

kenxiniz:
This adds multicast address that is different from the one of
resource offloading to avoid conflicts.

### DIFF
--- a/third_party/meerkat/Component/mmSH/android/server_runner_jni.cpp
+++ b/third_party/meerkat/Component/mmSH/android/server_runner_jni.cpp
@@ -19,6 +19,9 @@
 #include <android/log.h>
 #include <jni.h>
 
+#if defined(SERVICE_OFFLOADING)
+#include "base/distributed_chromium_util.h"
+#endif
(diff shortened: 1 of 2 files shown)


Chikdolman (approved):
LGTM

third_party/meerkat/Component/mmSH/android/server_runner_jni.cpp
```diff
@@ -169,7 +169,11 @@ jint Native_startServer(JNIEnv* env, jobject /* this */) {
 
   ServerRunner::ServerRunnerParams params;
   // TODO(yh106.jung): Read from configuration file
+#if defined(SERVICE_OFFLOADING)
```
Sunwoo-Nam, line 175: SERVICE_OFFLOADING flag is always enabled.
Please use a runtime flag.

kenxiniz: done

PR #370, opened 2020-11-18
Fix the crash in canvas2d rendering.

Sunwoo-Nam:
Skip CHECK() on |GetQueryObjectValueHelper|.
It causes a crash issue while using Gpu compositing
because the shared memory for query is not synchronized.

This patch is migrated from m69
https://github.com/Samsung/Castanets/pull/177


Chikdolman (approved):
Nice catch!

PR #374, opened 2020-11-24
[Service Offloading] Create the tab in incognito mode.

Sunwoo-Nam:
When creating a browser tab, the accessed url is cached.
Service offloading requires access to a specific url, but the problem may occur
if a different url is connected by other services.
To prevent that, create the tab in incognito mode so that it is not cached.

### DIFF
--- a/chrome/android/java/src/org/chromium/chrome/browser/ChromeTabbedActivity.java
+++ b/chrome/android/java/src/org/chromium/chrome/browser/ChromeTabbedActivity.java
@@ -1296,7 +1296,11 @@ public class ChromeTabbedActivity
                     "MobileStartup.LoadedHomepageOnColdStart", startupHomepageIsNtp);
         }
 
-        getTabCreator(false).launchUrl(url, TabLaunchType.FROM_CHROME_UI);
+        if (CommandLine.getInstance().hasSwitch(BaseSwitches.ENABLE_SERVICE_OFFLOADING)) {
+            getTabCreator(true).launchUrl(url, TabLaunchType.FROM_CHROME_UI);
+        } else {
+            getTabCreator(false).launchUrl(url, TabLaunchType.FROM_CHROME_UI);
+        }
     }
 
     @Override


chrome/android/java/src/org/chromium/chrome/browser/ChromeTabbedActivity.java
```diff
@@ -1296,7 +1296,11 @@ private void createInitialTab() {
                     "MobileStartup.LoadedHomepageOnColdStart", startupHomepageIsNtp);
         }
 
-        getTabCreator(false).launchUrl(url, TabLaunchType.FROM_CHROME_UI);
+        if (CommandLine.getInstance().hasSwitch(BaseSwitches.ENABLE_SERVICE_OFFLOADING)) {
+            getTabCreator(true).launchUrl(url, TabLaunchType.FROM_CHROME_UI);
+        } else {
+            getTabCreator(false).launchUrl(url, TabLaunchType.FROM_CHROME_UI);
+        }
```
Chikdolman, line 1303: To minimize code changes, how about attaching --disk-cache-size=1 or --incognito as a runtime option using meerkat?

Sunwoo-Nam: I hope runtime flags work with meerkat. When I tried it on the command line, it didn't work.

Sunwoo-Nam: And the incognito mode is always false in the initial tab regardless of the --incognito
> getTabCreator(false)

Chikdolman: @YoungsooChoi Can you check if --disk-cache-size=1 works with Meerkat?

kenxiniz: @Chikdolman 
I have tested with '--disk-cache-size=' but it's not working.

2020-11-24 15:21:41.621 16843-16843/com.samsung.android.castanets E/cr_ActivitiyDispatcher: YSYS ::: args = com.samsung.android.castanets --disk-cache-size=1 --enable-service-offloading --loadurl=file:///sdcard/Download/index.html

Chikdolman: Checked that --disk-cache-size did not affect.

Chikdolman (approved):

PR #375, opened 2020-11-24
[Meerkat] Remove unused switches for service offloading

kenxiniz:
This removes unused switches such as --enable-castanets and --server-address
for service offloading.

### DIFF
--- a/third_party/meerkat/Component/mmDiscovery/service_server.cpp
+++ b/third_party/meerkat/Component/mmDiscovery/service_server.cpp
@@ -110,18 +110,20 @@ VOID CServiceServer::DataRecv(OSAL_Socket_Handle iEventSock,
     std::vector<char*> argv;
     t_HandlePacket(argv, pData + strlen(kServiceRequestScheme));
 
-    char server_address[35] = {'\0',};
-    snprintf(server_address, sizeof(server_address) - 1,
-             "--enable-castanets=%s", pszsource_addr);
-    argv.push_back(server_address);
+    const std::string switches(argv.front());
+    if (switches.find("--enable-castanets")) {
+      char server_address[35] = {'\0',};
+      snprintf(server_address, sizeof(server_address) - 1,
+               "--enable-castanets=%s", pszsource_addr);
+      argv.push_back(server_address);
 
-    // TODO: |server_address_old| should be remove after applying
-    // https://github.com/Samsung/Castanets/pull/75.
-    char server_address_old[35] = {'\0',};
-    snprintf(server_address_old, sizeof(server_address_old) - 1,
-             "--server-address=%s", pszsource_addr);
-    argv.push_back(server_address_old);
-
+      // TODO: |server_address_old| should be remove after applying
+      // https://github.com/Samsung/Castanets/pull/75.
+      char server_address_old[35] = {'\0',};
+      snprintf(server_address_old, sizeof(server_address_old) - 1,
+               "--server-address=%s", pszsource_addr);
+      argv.push_back(server_address_old);
+    }
 #if defined(ANDROID)
     if (!Java_startCastanetsRenderer(argv))
 #else


third_party/meerkat/Component/mmDiscovery/service_server.cpp
```diff
@@ -110,18 +110,20 @@ VOID CServiceServer::DataRecv(OSAL_Socket_Handle iEventSock,
...
+    const std::string switches(argv.front());
+    if (!switches.find("--enable-service-offloading")) {
+      char server_address[35] = {'\0',};
+      snprintf(server_address, sizeof(server_address) - 1,
+               "--enable-castanets=%s", pszsource_addr);
+      argv.push_back(server_address);
+
+      // TODO: |server_address_old| should be remove after applying
+      // https://github.com/Samsung/Castanets/pull/75.
+      char server_address_old[35] = {'\0',};
```
Sunwoo-Nam, line 122: I don't think |server_address_old| is necessary.

third_party/meerkat/Component/mmDiscovery/service_server.cpp
```diff
@@ -110,18 +110,20 @@ VOID CServiceServer::DataRecv(OSAL_Socket_Handle iEventSock,
...
-
-    // TODO: |server_address_old| should be remove after applying
-    // https://github.com/Samsung/Castanets/pull/75.
-    char server_address_old[35] = {'\0',};
-    snprintf(server_address_old, sizeof(server_address_old) - 1,
-             "--server-address=%s", pszsource_addr);
-    argv.push_back(server_address_old);
-
+    const std::string switches(argv.front());
+    if (!switches.find("--enable-service-offloading")) {
```
Sunwoo-Nam: Why don't you check the condition like the below?
if (switches.find("--enable-castanets"))

third_party/meerkat/Component/mmDiscovery/service_server.cpp
kenxiniz: I just don't want to change original code because it seems to be used for resource offloading.

third_party/meerkat/Component/mmDiscovery/service_server.cpp
kenxiniz: The original code intended to add the --enable-castanets when it is not passed by 'ServiceRequest'. That's why following lines are used:

      char server_address[35] = {'\0',};
      snprintf(server_address, sizeof(server_address) - 1,
               "--enable-castanets=%s", pszsource_addr);
      argv.push_back(server_address);

kenxiniz: These are the arguments from resource offloading and there's no --enable-castanets but they need it.

I/CHROMIUM( 3553): [INFO:WRT(0)] app.js RequestService : castanets, android, com.samsung.android.castanets --type=offloadworker --signaling-server=http://192.168.0.103:9559/offload-js, 192.168.0.96

third_party/meerkat/Component/mmDiscovery/service_server.cpp
kenxiniz: I asked @younghajung and noticed that the added address is needed for renderer offloading on castanets.

Sunwoo-Nam: The address is appended with --enable-castanets.

third_party/meerkat/Component/mmDiscovery/service_server.cpp
Sunwoo-Nam: Do you mean resource offloading needs --enable-castanets flag?

kenxiniz: no, @younghajung told me only render offloading uses the flag. we can use --enable-castanets flag in the if condition. I think the --enable-castanets flag has been added to resource offloading but it was ineffective for the resource offloading.

Sunwoo-Nam (approved):
Looks good to me

PR #376, opened 2020-11-24
[Service Offloading] Don't display https certificate warning

kenxiniz:
The service offloading scenario doesn't have webview to show
and users are noticed that their app screen will be shared
with Tizen devices. So, the https certificate warning is not necessary.

### DIFF
--- a/content/browser/android/content_startup_flags.cc
+++ b/content/browser/android/content_startup_flags.cc
@@ -27,6 +27,7 @@
 
 #if defined(CASTANETS) || defined(SERVICE_OFFLOADING)
 #include "base/distributed_chromium_util.h"
+#include "components/network_session_configurator/common/network_switches.h"
 #endif
 
 namespace content {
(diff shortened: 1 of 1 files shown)


Chikdolman (approved):
LGTM, but how about to append switch to base/base_switches instead of using "ignore-certificate-errors" directly?

content/browser/android/content_startup_flags.cc
```diff
@@ -69,6 +69,12 @@ void SetContentCommandLineFlags(bool single_process) {
     base::CommandLine::ForCurrentProcess()->AppendSwitchASCII(
         switches::kDisableFeatures, disabled_features);
   }
+#endif
+#if defined(SERVICE_OFFLOADING)
+  if (base::ServiceOffloading::IsEnabled()) {
```
Sunwoo-Nam: I think it is better to move this code to the bottom of the function and integrate with the same condition.

kenxiniz:
> LGTM, but how about to append switch to base/base_switches instead of using "ignore-certificate-errors" directly?

Done

content/browser/android/content_startup_flags.cc
kenxiniz: Done

PR #379, opened 2020-12-01
[Meerkat] Use ini file for the initialization on debug mode

younghajung:
This patch uses an ini file for the initialization of Meerkat server
when Service Offloading is set as debug app.

1. Settings > Developer options > Select debug app
2. In the terminal, adb push [your ini fle] /data/local/tmp/meerkat/server.ini
3. Start Meerkat


is46-kim (approved):
LGTM

PR #380, opened 2020-12-02
[Service Offloading] Support Multitouch on Android

Chikdolman:
Since we only support mouse event, FHUB and kiosk are not supported.
Therefore we need to support multitouch.
We enable multi-touch with this patch.

### DIFF
--- a/third_party/blink/renderer/modules/input_control/InputControl.java
+++ b/third_party/blink/renderer/modules/input_control/InputControl.java
@@ -16,6 +16,8 @@ import android.net.wifi.WifiInfo;
 import android.net.wifi.WifiManager;
 import android.view.KeyEvent;
 import android.view.MotionEvent;
+import android.view.MotionEvent.PointerCoords;
+import android.view.MotionEvent.PointerProperties;
 import android.widget.TextView;
 
 import com.samsung.android.knox.EnterpriseDeviceManager;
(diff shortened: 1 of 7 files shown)


third_party/blink/renderer/modules/input_control/InputControl.java
```diff
@@ -58,7 +70,50 @@ private static InputControl CreateInputControl() {
...
+    }
+
+    @CalledByNative
+    public void SendTouchInput(String type, String json) {
+        EnterpriseDeviceManager edm = EnterpriseDeviceManager.getInstance(
+          ContextUtils.getApplicationContext());
+        RemoteInjection remoteInjection = edm.getRemoteInjection();
+        int action = -1;
+
+        if (type.equals("touchstart") && mNumTouch < 1) {
```
Sunwoo-Nam: How about this?
if (type.equals("touchstart")) {
  if (mNumTouch < 1) {
    ....
  } else {
    .....
  }
} else if (type.equals("touchend")) {
  ....
} else if (type.equals("touchmove")) {
  ....
}

Chikdolman: Thank you 👍 applied

Sunwoo-Nam (approved):
Please merge if the behavior is verified.

kenxiniz (approved):
lgtm

PR #381, opened 2020-12-02
[Meerkat] Append browser's address only when castanets enabled

younghajung:
This patch appends browser's address to castanets switch only when
castanets enabled.


is46-kim (approved):
LGTM

PR #385, opened 2020-12-10
Load meerkat library with ".cr" suffix for component build

is46-kim:
In a component build, the ".cr" suffix is added to each library name.
If the meerkat library loading fails, try to load the library with ".cr"
suffix for component build.


younghajung (approved):
LGTM

PR #386, opened 2020-12-15
[Service Offloading] Allow getDisplayMedia in non-secure context.

Sunwoo-Nam:
In Service offloading, the android device gets contents for the webrtc server
from the client device. The android device needs to use http protocol to access
the client device, but getDisplayMedia does not work with http content.
So, this patch allows that getDisplayMedia works on the http content by setting
the document state to 'Secure'.


Sunwoo-Nam:
Reference: https://github.sec.samsung.net/k2-nagaraju/WebRTCServices/commit/9523389f7b409c93ee2ca979d156ff96ad07d2c4

Chikdolman (approved):
LGTM. However, the patch should be merged after verification.

PR #387, opened 2020-12-16
Update Settings and Remove App Icon

is46-kim:
- Remove Castanets App Icon for testing in Launcher
- Replace settings MainActivity with SettingsActivity
- Start Meerkat service only when Castanets is enabled


younghajung (approved):
LGTM

PR #391, opened 2021-02-05
Add OffloadPermissionActivity to AndroidManifest

djmixkim:
We added a new acivitiy as OffloadPermissionActivity for permissions UI
in offload.js
This patch registers information of activity to manifest.


is46-kim (approved):
LGTM

PR #399, opened 2021-04-26
Use strtok_r() instead of strtok() for thread safety

is46-kim:


HunseopJeong (approved):
LGTM

PR #401, opened 2021-05-20
[Meerkat] Launch OffloadService directly

is46-kim:
This starts OffloadService directly, not through Castanets.

### DIFF
--- a/third_party/meerkat/Component/mmSH/android/java/src/com/samsung/android/meerkat/MeerkatServerService.java
+++ b/third_party/meerkat/Component/mmSH/android/java/src/com/samsung/android/meerkat/MeerkatServerService.java
@@ -39,7 +39,11 @@ import android.support.v4.app.NotificationCompat.Builder;
 import android.util.Log;
 
 import java.io.File;
+import java.lang.reflect.Method;
 import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
 import org.json.simple.JSONObject;
 import org.json.simple.parser.JSONParser;
 import org.json.simple.parser.ParseException;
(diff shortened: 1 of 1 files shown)


third_party/meerkat/Component/mmSH/android/java/src/com/samsung/android/meerkat/MeerkatServerService.java
```diff
@@ -196,6 +200,15 @@ private static boolean shouldUseDebugIni() {
     }
 
     public static boolean startCastanetsRenderer(String args) {
+        // Launch OffloadService directly.
+        if (args.matches("(^|.*\\s)--type=offloadworker($|\\s.*)")) {
```
HunseopJeong, line 204: Is the '\\s' correct regex?
Isn't the '\s' enough to check if there is space before and after?

is46-kim: > Is the '\s' correct regex?
> Isn't the '\s' enough to check if there is space before and after?

No, It is not matched only with "\s" without ". *".
I also finally found that ". *" needed to check a space before and after.

HunseopJeong: Yes, I mean, Isn't "(^|.*\s) .... ($|\s.*)" enough? instead of the '\\\s'

is46-kim: Java escape character is backslash('\'). So it needs double backslash for backslash, not escape.

HunseopJeong: I see. Thank you for your kind explanation.

HunseopJeong (approved):
LGTM

PR #425, opened 2021-10-08
Fix Tizen-Android interoperability issues.

suyambulingamrm:
This change makes IPC params between tizen and android similar so that
the IPC messages sent from android are recognized properly.


content/public/common/untrustworthy_context_menu_params.h
```diff
@@ -117,6 +117,12 @@ struct CONTENT_EXPORT UntrustworthyContextMenuParams {
   // Whether context is editable.
   bool is_editable;
 
+#if defined(CASTANETS)
+  bool is_text_node;
+  bool is_draggable;
+  bool is_user_select_none;
```
younghajung: Don't we need initialization for these?

suyambulingamrm: I have initialized it with default values in latest patchset.
I added just to match the params, It didn't serve any logical purpose for now.

younghajung (approved):
LGTM

